PR #1, opened 2016-02-25
created initial migration and models

fillup:
@forevermatt @Baggerone


application/common/helpers/Utils.php
```diff
@@ -0,0 +1,88 @@
...
+namespace common\helpers;
+
+use yii\base\Security;
+
+class Utils
+{
+
+    const DT_Format = 'Y-m-d H:i:s';
+
+    public static function getDatetime($timestamp=false)
```
forevermatt: Can we change `$timestamp=false` to `$timestamp = null` to match our use of null (not false) as our "no data" value?  (Similarly for `getIso8601(...)` below.)

application/common/models/PasswordChangeLogBase.php
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * @inheritdoc
+     */
+    public function rules()
+    {
+        return [
+            [['user_id', 'scenario', 'reset_type', 'method_type', 'masked_value', 'created', 'ip_address'], 'required'],
+            [['user_id'], 'integer'],
+            [['scenario', 'reset_type', 'method_type'], 'string'],
+            [['created'], 'safe'],
```
forevermatt: I'm not sure if there's a simple way to change this, but all of these `created` columns should probably only be `safe` on insert (not update), since we wouldn't want the created value to be changed.

application/common/models/PasswordChangeLogBase.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        return [
+            'id' => Yii::t('app', 'ID'),
+            'user_id' => Yii::t('app', 'User ID'),
+            'scenario' => Yii::t('app', 'Scenario'),
+            'reset_type' => Yii::t('app', 'Reset Type'),
+            'method_type' => Yii::t('app', 'Method Type'),
+            'masked_value' => Yii::t('app', 'Masked Value'),
+            'created' => Yii::t('app', 'Created'),
+            'ip_address' => Yii::t('app', 'Ip Address'),
```
forevermatt: Very minor issue, but if you wanted you could override `attributeLabels()` in the subclass to change "Ip Address" to "IP Address".

application/common/models/Reset.php
```diff
@@ -0,0 +1,47 @@
...
+ * Class Reset
+ * @package common\models
+ * @method Reset self::findOne([])
+ */
+class Reset extends ResetBase
+{
+
+    const TYPE_METHOD = 'method';
+    const TYPE_SUPERVISOR = 'supervisor';
+    const TYPE_SPOUSE = 'SPOUSE';
```
forevermatt: Was this intentionally in all caps?

application/common/models/User.php
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * Finds an identity by the given token.
+     * This method is not supported in this app right now but is required by Yii IdentityInterface
+     *
+     * @param string $token the token to be looked for
+     * @return IdentityInterface|null the identity object that matches the given token.
+     */
+    public static function findIdentityByAccessToken($token, $type = null)
+    {
+        return static::findOne($token);
```
forevermatt: Should this return `null` (since we're not supporting this method at the moment) instead of using `$token` as though it were an identity's `$id`?

application/common/models/UserBase.php
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public function getPasswordChangeLogs()
+    {
+        return $this->hasMany(PasswordChangeLog::className(), ['user_id' => 'id']);
+    }
+
+    /**
+     * @return \yii\db\ActiveQuery
+     */
+    public function getReset()
```
forevermatt: It looks like Yii (incorrectly?) assumed that since PasswordChangeLogs and Resets both have a foreign key to User, they therefore have a relationship to each other.  But there's not any actual guarantee of a relationship between a particular PasswordChangeLog entry and the current Reset (if any) that has a matching `user_id`, right?

application/composer.json
```diff
@@ -1,26 +1,27 @@
...
-    "extra": {
-        "asset-installer-paths": {
-            "npm-asset-library": "vendor/npm",
-            "bower-asset-library": "vendor/bower"
-        }
+  "minimum-stability": "stable",
+  "require": {
+    "php": ">=5.4.0",
+    "yiisoft/yii2": "2.0.6",
+    "yiisoft/yii2-bootstrap": "*",
```
forevermatt: Would it be a good idea to change our `require` version strings from `*` to instead use the `^some.specific.version` syntax?

application/console/migrations/m160225_162828_create_initial_tables.php
```diff
@@ -0,0 +1,132 @@
...
+         * Method table
+         */
+        $this->createTable(
+            '{{method}}',
+            [
+                'id' => 'pk',
+                'user_id' => 'int(11) not null',
+                'type' => "enum('email','phone') not null",
+                'value' => 'varchar(255) not null',
+                'verified' => 'boolean not null',
```
forevermatt: Will we need to specify that `verified` should have a default value of `false`?

application/common/helpers/Utils.php
fillup: Yep, will fix.

application/common/models/Reset.php
fillup: nope, will fix.

application/common/models/User.php
fillup: Sure. will fix.

application/console/migrations/m160225_162828_create_initial_tables.php
fillup: Sure, will add it.

PR #7, opened 2016-02-29
Routes defined, models and controllers created, unit tests set up

fillup:
This branch turned in to a lot more than I originally intended, but includes routes defined, controllers created, some of the actions working, models in place, some working, etc. I want to get this back into develop so other branches for ongoing work can continue before this one gets any bigger.

@forevermatt @wcjr @Baggerone


fillup:
Also, codeship is having problems so ignore the failed build there.

wcjr:
lgtm

README.md
```diff
@@ -25,10 +25,16 @@ Backend API for Identity Provider Password Management
 ## Setup
 1. Clone this repo
 2. Copy ```local.env.dist``` to ```local.env``` and update values as appropriate
-3. Follow operating system specific steps below
-4. You should be able to access the API using a REST client or your browser
+3. Setup environment variable for ```DOCKER_UIDGID``` in the format of ```"uid:gid"```.
+   This will run some of the containers as you so that they can write to your host filesystem
+   and the file permissions will be owned you you.
```
forevermatt: Extra "you" in here.

application/common/models/PasswordChangeLog.php
```diff
@@ -25,21 +25,21 @@ public function rules()
...
+                    'message' => 'Scenario must be either ' . self::SCENARIO_CHANGE . ' or ' .
+                        self::SCENARIO_RESET . '.',
                 ],
 
                 [
                     ['reset_type'], 'in', 'range' => [Reset::TYPE_METHOD, Reset::TYPE_SUPERVISOR, Reset::TYPE_SPOUSE],
-                    'message' => 'Reset type must be either '.Reset::TYPE_METHOD.' or '.
-                        Reset::TYPE_SUPERVISOR.' or '.Reset::TYPE_SPOUSE.' .',
+                    'message' => 'Reset type must be either ' . Reset::TYPE_METHOD . ' or ' .
+                        Reset::TYPE_SUPERVISOR . ' or ' . Reset::TYPE_SPOUSE . ' .',
```
forevermatt: Looks like we'll need to add Reset::TYPE_PRIMARY to this list.

application/common/models/Reset.php
```diff
@@ -33,9 +40,19 @@ public function rules()
...
+                [
+                    ['attempts'], 'default', 'value' => 0,
+                ],
+
+                [
+                    ['type'], 'in', 'range' => [
+                        self::TYPE_PRIMARY, self::TYPE_METHOD, self::TYPE_SUPERVISOR, self::TYPE_SPOUSE
+                    ],
+                    'message' => 'Reset type must be either ' . self::TYPE_METHOD . ' or ' .
+                        self::TYPE_SUPERVISOR . ' or ' . self::TYPE_SPOUSE . ' .',
```
forevermatt: May similarly need to add Reset::TYPE_PRIMARY here.

application/common/models/Reset.php
```diff
@@ -43,14 +60,160 @@ public function rules()
...
+     * @return Reset
+     * @throws NotFoundHttpException
+     * @throws \Exception
+     */
+    public static function findOrCreate($user, $type = self::TYPE_PRIMARY, $methodId = null)
+    {
+        /*
+         * Find existing or create new Reset
+         */
+        $reset = self::findOne(['user_id' => $user->id]);
```
forevermatt: Could we use `$user->reset` here?

application/common/models/Reset.php
```diff
@@ -43,14 +60,160 @@ public function rules()
...
+             */
+            if ( ! $reset->save()) {
+                throw new \Exception('Unable to create new reset', 1456608028);
+            }
+        }
+
+        /*
+         * Send Reset to appropriate method
+         */
+        $reset->send();
```
forevermatt: Should we be calling `send()` even if the Reset already existed? I would have guess from the name `findOrCreate(...)` that it would also automatically send the reset code.

application/common/models/Reset.php
```diff
@@ -43,14 +60,160 @@ public function rules()
...
+                $this->sendSupervisor();
+                break;
+            case self::TYPE_SPOUSE:
+                $this->sendSpouse();
+                break;
+            case self::TYPE_METHOD:
+                $this->sendMethod();
+                break;
+            default:
+                $this->sendPrimary();
```
forevermatt: This seems like a good place to throw an exception (rather than just sending to primary), since the type is apparently set to something that we aren't yet prepared to handle. Thoughts?

application/common/models/Reset.php
```diff
@@ -43,14 +60,160 @@ public function rules()
...
+    {
+        /**
+         * @todo send email to $user->email with reset code
+         */
+    }
+
+    public function sendSupervisor()
+    {
+        /**
+         * @todo if $this->user->getHasSupervisor(), send reset
```
forevermatt: Not that it affects anything yet, but `getHasSupervisor()` looks like it should just be `hasSupervisor()`.

application/common/models/Reset.php
```diff
@@ -43,14 +60,160 @@ public function rules()
...
+         *       code to $this->user->getSupervisorEmail()
+         */
+    }
+
+    public function sendSpouse()
+    {
+        /**
+         * @todo if $this->user->getHasSpouse(), send reset
```
forevermatt: Similarly, `getHasSpouse()` -> `hasSpouse()`.

application/common/models/ResetBase.php
```diff
@@ -36,11 +37,13 @@ public static function tableName()
     public function rules()
     {
         return [
-            [['user_id', 'type', 'code', 'expires', 'created'], 'required'],
```
forevermatt: That's interesting. Any idea why it no longer considers `type` a required field?  My only guess is that it sees a default value was provided in the column definition (in the migration file).

application/common/models/Reset.php
fillup: yep. fixed.

application/common/models/Reset.php
fillup: fixed.

application/common/models/PasswordChangeLog.php
fillup: fixed.

application/common/models/User.php
```diff
@@ -40,6 +48,195 @@ public function rules()
...
+        if (substr_count($username, '@') > 0) {
+            $personnelUser = \Yii::$app->personnel->findByEmail($username);
+        } else {
+            $personnelUser = \Yii::$app->personnel->findByUsername($username);
+        }
+
+        $user = self::findOne(['employee_id' => $personnelUser->employeeId]);
+        if ( ! $user) {
+            $user = new User();
+            $user->employee_id = $personnelUser->employeeId;
```
forevermatt: Is it possible for `$personnelUser` here to be `null` (such as if no such user was found in the personnel system)?

README.md
fillup: fixed.

application/common/models/Reset.php
fillup: I was thinking yes since the scenario is the user never completed the initial reset and are "starting" over. I expect send() to enforce any rules about frequency of sending and limits and what not to prevent spam.

application/common/models/Reset.php
fillup: sure, fixed.

application/common/models/Reset.php
fillup: fixed.

application/common/models/Reset.php
fillup: fixed.

application/common/models/ResetBase.php
fillup: yeah that is my guess too, maybe we should remove the default from the migration since it is implemented in the model too.

application/common/models/User.php
```diff
@@ -40,6 +48,195 @@ public function rules()
...
+    }
+
+
+    /**
+     * Update local user record if given properties are different than currently stored
+     * @param string $firstName
+     * @param string $lastName
+     * @param string $username
+     * @param string $email
+     * @return bool True if profile was updated, false if not
```
forevermatt: It might be worth clarifying that `false` here means it wasn't updated because it didn't need to be updated (it all matched). As it is, someone could misread it as "it tried to update and failed" (such as if a validation rule failed).

application/common/models/Reset.php
forevermatt: I would still lean towards moving that into some other part of the code.  If we need to re-send a code when someone restarts the process, that's fine.  The `findOrCreate` function doesn't seem like the right place to do that, though. Just FWIW.

application/common/models/Reset.php
```diff
@@ -43,14 +60,159 @@ public function rules()
...
+     * @param null|integer $time
      * @return integer
      */
-    public static function getExpireTimestamp($time=null)
+    public static function getExpireTimestamp($time = null)
     {
         $time = is_null($time) ? time() : $time;
 
-        return $time+\Yii::$app->params['reset']['lifetimeSeconds'];
+        return $time + \Yii::$app->params['reset']['lifetimeSeconds'];
```
forevermatt: Would it be worth adding a sanity check here to make sure the resulting value isn't too far in the past or the future (or at least ensuring that the result is an integer)? Just a thought.

application/common/models/User.php
fillup: No, it should thrown an exception if user is not found in personnel.

application/common/models/User.php
```diff
@@ -40,6 +48,195 @@ public function rules()
...
+    {
+        if ( ! empty($this->personnelUser)) {
+            return $this->personnelUser;
+        }
+
+        /*
+         * If session is available, cache data in session
+         */
+        try {
+            $sessionAvailable = (\Yii::$app->user->identity == $this);
```
forevermatt: Is this logic something that should be pulled out into a separate function? It seems non-intuitive enough that it might be worth having some unit tests to prove that it does what it says. Doing so could also improve the readability of this code.

application/common/models/User.php
```diff
@@ -40,6 +48,195 @@ public function rules()
...
+        } catch (\Exception $e) {
+            $sessionAvailable = false;
+        }
+
+        if ($sessionAvailable && is_array(\Yii::$app->session->get('personnelUser'))) {
+            return \Yii::$app->session->get('personnelUser');
```
forevermatt: Before returning the personnelUser data array here, should it be (1) converted into an actual PersonnelUser object and/or (2) stored back into `$this->personnelUser`?

application/common/models/User.php
```diff
@@ -40,6 +48,195 @@ public function rules()
...
+        return $personnelUser->supervisorEmail !== null;
+    }
+
+    /**
+     * @return bool
+     */
+    public function hasSpouse()
+    {
+        $personnelUser = $this->getPersonnelUser();
+        return $personnelUser->spouseEmail !== null;
```
forevermatt: Not that it's much code, but we could slightly reduce duplication here by having this `hasSpouse()` function use the `getSpouseEmail()` function internally, since they do almost exactly the same thing.  Same with the supervisor functions here.  Just a thought.

application/composer.json
```diff
@@ -5,7 +5,9 @@
     "yiisoft/yii2": "2.0.6",
     "yiisoft/yii2-bootstrap": "*",
     "yiisoft/yii2-swiftmailer": "*",
-    "silinternational/php-env": "^0.1.1"
+    "silinternational/php-env": "^0.1.1",
+    "silinternational/yii2-jsonsyslog": "*",
+    "silinternational/idp-pw-api-common": "dev-master"
```
forevermatt: Is there a difference between `dev-master` and `*`?

application/frontend/config/main.php
```diff
@@ -2,25 +2,20 @@
 
 /* Get frontend-specific config settings from ENV vars or set defaults. */
 $frontCookieKey = getenv('FRONT_COOKIE_KEY') ?: null;
+const UID_ROUTE_PATTERN = '<uid:([a-zA-Z0-9_\-]{32})>';
```
forevermatt: As things currently stand, this matches the values we're using for UIDs. Is there some way we can make sure we know to update this if we ever change the possible format of a UID?  (e.g. - Should/could we add a unit test that simply checks ten values returned by our UID-generating code against this pattern [retrieved from the same place this copy of it is retrieved from]?).

I guess I'm suggesting that some static methods be added in a single class that handles this regex and the generation of our UIDs, and using those static methods anywhere we need either this regex or a new UID.

application/frontend/controllers/MethodController.php
```diff
@@ -0,0 +1,49 @@
...
+                'class' => AccessControl::className(),
+                'rules' => [
+                    [
+                        'allow' => true,
+                        'roles' => ['@'],
+                    ],
+                    [
+                        'allow' => true,
+                        'actions' => ['index', 'view'],
+                        'roles' => ['?'],
```
forevermatt: Is this saying that unauthenticated users can view a method's information?

application/frontend/controllers/MethodController.php
```diff
@@ -0,0 +1,49 @@
...
+                    ],
+                ]
+            ]
+        ]);
+    }
+
+    /**
+     * Return list of available reset methods for user.
+     * If user is not authenticated they should be masked.
```
forevermatt: Is masking these still a todo item, or is that already being handled somewhere else?  (This looked like a `@todo` comment, but since it lacked the tag it won't be flagged in Scrutinizer.)

application/frontend/controllers/ResetController.php
```diff
@@ -0,0 +1,93 @@
...
+use yii\filters\AccessControl;
+use yii\helpers\ArrayHelper;
+use yii\web\BadRequestHttpException;
+use yii\web\NotFoundHttpException;
+
+class ResetController extends BaseRestController
+{
+    /**
+     * Access Control Filter
+     * NEEDS TO BE UPDATED FOR EVERY ACTION
```
forevermatt: Come to think of it, should these "NEEDS TO..." comments have a `@todo` stuck on them?

application/tests/codeception/config/functional.php
```diff
@@ -1,18 +1,24 @@
 <?php
 /**
- * Application configuration shared by all applications functional tests
+ * Application configuration shared by all applications unit tests
```
forevermatt: Looks like this may have been an accidental change, since this is the `functional.php` config file.

application/tests/unit/common/models/MethodTest.php
```diff
@@ -0,0 +1,61 @@
...
+        $this->assertNull($method->verification_code);
+        $this->assertEquals(0, $method->verification_attempts);
+        $this->assertNull($method->verification_expires);
+        $this->assertNotNull($method->created);
+    }
+
+    public function testGetMaskedValuePhone()
+    {
+        $method = $this->methods('method1');
+        $this->assertEquals('1,1234567890', $method->getMaskedValue());
```
forevermatt: Is it just me, or does this test prove that `getMaskedValue()` *doesn't* work? It seems to be asserting that the returned value is not masked. If this is just because we haven't implemented that masking logic yet, it might be better to assert that they're not equal, then mark this test as skipped for now.

application/tests/unit/common/models/MethodTest.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public function testGetMaskedValueEmail()
+    {
+        $method = $this->methods('method2');
+        $this->assertEquals('email-1456769679@domain.org', $method->getMaskedValue());
```
forevermatt: Same comments as for `testGetMaskedValuePhone()`.

local.env.dist
```diff
@@ -14,3 +14,7 @@ SAML_CERT_FINGERPRINT=
 SAML_ORG_NAME=
 SAML_ORG_URL=
 FRONT_COOKIE_KEY=akljwelkjfwlkajwfkljeafewf5rgtw2
+RECAPTCHA_SITE_KEY=
+RECAPTCHA_SECRET_KEY=
+COMPOSER_AUTH={"github-oauth":{"github.com":"tokenhere"}
```
forevermatt: Looks like a closing `}` is missing here.

forevermatt:
Okay @fillup. Once you feel like all of the (many) things I commented have been sufficiently addressed, you're free to merge this PR as far as I'm concerned.

application/common/models/User.php
fillup: I agree, I copied this pattern over from doorman but I'm sure there is a better way to check this and it would make sense to move to a separate location.

Baggerone:
For `docker-compose up -d` to work, I needed to add this to the last part of the Vagrantfile ...
`export COMPOSER_CONFIG_FILE=/vagrant/application/composer.json`

Baggerone:
Does the ui_url param need to be set in one of the config files? 
/data/frontend/controllers/SiteController.php(49)

application/common/models/User.php
fillup: Updated.

application/common/models/Reset.php
fillup: I removed the call to send, we will expect controller or calling function to also call send if needed.

application/common/models/ResetBase.php
fillup: I removed the default from the migration so it only exists in our extended model.

application/common/models/User.php
fillup: Moved to Utils

application/common/models/User.php
fillup: Assume part of deserialization, will test when we get there though

application/common/models/User.php
fillup: good call

application/frontend/controllers/MethodController.php
fillup: Yes, but we've changed that plan so I'll remove that.

application/frontend/controllers/MethodController.php
fillup: fixed

PR #9, opened 2016-04-01
add personnel component and masking

fillup:
I've added a new ```local.php``` config file that will need to be injected into each environment, but it should allow greater flexibility in providing local specific configurations for various components like personnel, sms, etc. There are ```dist``` and ```test``` versions of the file with ```dist``` to be a starting point and ```test``` to hold whatever is needed for unit tests.

I also added the ```personnel``` component to the configuration and am including the first personnel component defined so far ```idp-pw-api-personnel-insite```. 

Finally, I copied over the masking methods from another project to mask email addresses and phone numbers.

@forevermatt @Baggerone @wcjr


Baggerone:
Looks good to me. I don't see a link for approving the PR.

PR #11, opened 2016-04-07
saml auth component

fillup:
Integrated idp-pw-api-auth-saml for login/logout

@forevermatt @Baggerone @wcjr


application/common/models/User.php
```diff
@@ -48,29 +49,59 @@ public function rules()
...
+     * Limit what fields are returned from api calls
+     * @return array
+     */
+    public function fields()
+    {
+        return [
+            'first_name',
+            'last_name',
+            'idp_username' => function($model) {
+                return strtolower($model->idp_username);
```
forevermatt: What's the reasoning behind lowercasing the idp_username?  Given the IdP we initially plan to use this with, I think that will be fine.  I just wasn't sure if that's generally true of IdP usernames, or if we would expect some to be case-sensitive.

application/common/models/User.php
```diff
@@ -278,4 +309,20 @@ public function validateAuthKey($authKey)
...
+
+    /**
+     * Get this user as an AuthUser object
+     * @return AuthUser
+     */
+    public function getAuthUser()
+    {
+        $authUser = new AuthUser();
+        $authUser->firstName = $this->first_name;
+        $authUser->lastName = $this->last_login;
```
forevermatt: It looks like autocomplete caused a mistake here.  My guess is that you wanted last_name, not last_login.

application/common/models/User.php
```diff
@@ -278,4 +309,20 @@ public function validateAuthKey($authKey)
     {
         return false;
     }
+
+    /**
+     * Get this user as an AuthUser object
+     * @return AuthUser
```
forevermatt: Should the @return parameter use the actual class name/path (\Sil\IdpPw\Common\Auth\User) rather than the "use" alias we're using in this file?

application/frontend/controllers/AuthController.php
```diff
@@ -0,0 +1,123 @@
...
+        $log = ['action' => 'login'];
+
+        try {
+            $returnTo = Url::to(['auth/login', 'ReturnTo' => $returnTo], true);
+            /** @var AuthUser $authUser */
+            $authUser = \Yii::$app->auth->login(\Yii::$app->request, $returnTo);
+
+            $log['email'] = $authUser->email;
+            // Get local user instance or create one
+            $user = User::findOrCreate(null, null, $authUser->employeeId);
```
forevermatt: We're intentionally only using the `employeeId` when finding the user's record when they log in here, right? I wonder if it might be worth adding a comment to remind us of that fact, so that we don't later replace these `null` values with attributes of the `$authUser`.

application/common/models/User.php
fillup: Yeah good point, I was mainly thinking that this data is returned to the UI/client and would be displayed and being all caps is ugly. But I could leave it as is and let the UI/client format it however they want.

application/frontend/controllers/AuthController.php
```diff
@@ -0,0 +1,123 @@
...
+        /*
+         * Kill local session
+         */
+        \Yii::$app->user->logout(true);
+
+        /*
+         * Log user out of IdP
+         */
+        try {
+            \Yii::$app->auth->logout($authUser, \Yii::$app->params['ui_url']);
```
forevermatt: If the call to `\Yii::$app->auth->logout(...)` doesn't throw a `RedirectException`, what would happen?  Should we instead just be getting the URL from the `RedirectException` (if thrown, otherwise using the ui_url), and then waiting until after the `catch` block to do the redirect?

application/common/models/User.php
fillup: Yup, good catch!

application/common/models/User.php
fillup: sure.

application/frontend/controllers/AuthController.php
```diff
@@ -0,0 +1,123 @@
...
+        } catch (RedirectException $e) {
+            return $this->redirect($e->getUrl());
+        }
+
+    }
+
+    public function getAfterLoginUrl($returnTo)
+    {
+        /*
+         * Only keep $returnTo if it is a path on the frontend
```
forevermatt: Do you think it would be worth adding something to this comment to remind us (down the road) why we only allow redirecting to relative paths?  (i.e. - to prevent phishing attacks using the actual login process to send someone to a bad/fake website, etc.)

application/frontend/controllers/AuthController.php
fillup: sure.

application/tests/codeception/common/templates/fixtures/user.php
```diff
@@ -1,8 +1,8 @@
 <?php
 /**
- * @var $faker \Faker\Generator
- * @var $index integer
- */
+     * @var $faker \Faker\Generator
```
forevermatt: I would suggest setting these Codeception files' comments back to how they were. We apparently have something in our scrutinizer config that encourages Scrutinizer to mess up the indentation here.

local.env.dist
```diff
@@ -2,17 +2,6 @@
...
-SAML_ADMIN_PASSWORD=
-SAML_SECRET_SALT=
-SAML_CONTACT_NAME=
-SAML_CONTACT_EMAIL=
-SAML_SSO_URL=
-SAML_SLO_URL=
-SAML_CERT_FINGERPRINT=
-SAML_ORG_NAME=
-SAML_ORG_URL=
 FRONT_COOKIE_KEY=akljwelkjfwlkajwfkljeafewf5rgtw2
```
forevermatt: Would it be worth changing this dummy value to something more likely to get replaced (such as 'aaaaaa' or 'replace-me-with-a-secure-random-string')? Just a thought.

application/frontend/controllers/AuthController.php
```diff
@@ -0,0 +1,123 @@
...
+            $log['code'] = $e->getCode();
+            \Yii::error($log, 'application');
+            throw new UnauthorizedHttpException($e->getMessage(), $e->getCode());
+        }
+
+    }
+
+    public function actionLogout()
+    {
+        if ( ! \Yii::$app->user->isGuest) {
```
Baggerone: How about getting rid of the else by saying `if (\Yii::$app->user->isGuest) { \Yii::$app->user->logout(true); return ...}`

application/frontend/controllers/AuthController.php
fillup: yeah I'll add a redirect to ui_url after the catch

application/frontend/controllers/AuthController.php
fillup: ok

application/tests/codeception/common/templates/fixtures/user.php
fillup: that is why I finally accepted them. Since these are files auto-generated by codeception and scrutinizer kept wanting to indent them, and even ignoring the suggestion kept it on the list so I just accepted to get them out of the way since they dont actually break anything.

local.env.dist
fillup: ok

application/frontend/controllers/AuthController.php
fillup: ok

PR #16, opened 2016-04-12
Integrated phone component

fillup:
@Baggerone @forevermatt @wcjr


application/common/models/Reset.php
```diff
@@ -146,23 +146,23 @@ public function send()
 
     public function sendPrimary()
     {
-        /**
-         * @todo send email to $user->email with reset code
+        /*
+         * send email to $user->email with reset code
```
Baggerone: I don't understand why you removed the @todo's

application/common/models/Reset.php
```diff
@@ -188,16 +188,59 @@ public function sendMethod()
...
+        ];
+
+        // Get phone number without comma
+        $number = $this->method->getRawPhoneNumber();
+
+        // Generate random code for potential use
+        $code = Utils::getRandomDigits(\Yii::$app->phone->codeLength);
+
+        // Call component send() method to send verification and capture resulting code
+        $result = \Yii::$app->phone->send($number, $code);
```
Baggerone: Should the app really be mixing together configs with method calls, i.e. \Yii:$app->phone->codeLength vs. \Yii::$app->phone->send()?

application/tests/mock/auth/Component.php
```diff
@@ -0,0 +1,82 @@
...
+        /*
+         * Loop through mock users to perform "login"
+         */
+        $data = include __DIR__ . '/data.php';
+        foreach ($data as $user) {
+            if ($user['username'] == $username) {
+                if ($user['password'] == $password) {
+                    return $this->toAuthUser($user);
+                } else {
+                    break;
```
Baggerone: You don't need the `else`, just the `break;`

application/tests/unit/common/models/ResetTest.php
```diff
@@ -94,5 +94,18 @@ public function testFindOrCreateExisting()
...
     }
 
+    public function testSendPhone()
+    {
+        $reset = $this->resets('reset2');
+        $this->assertNull($reset->code);
+        $reset->send();
+        $this->assertEquals('1234', $reset->code);
+        $this->assertEquals(1, $reset->attempts);
+        $this->assertTrue($reset->isUserProvidedCodeCorrect('1234'));
```
Baggerone: Normally it's better to have separate tests for testing the different methods.

application/common/models/Reset.php
fillup: I created issues in github instead, the @todo impacts the score in scrutinizer and they aren't a very good way to keep track of work needed anyway

application/tests/mock/auth/Component.php
fillup: fixed

application/tests/unit/common/models/ResetTest.php
fillup: fixed

application/common/models/Reset.php
```diff
@@ -146,24 +146,22 @@ public function send()
 
     public function sendPrimary()
     {
-        /**
-         * @todo send email to $user->email with reset code
```
forevermatt: @fillup, just checking: why did we remove the `@todo` flags before the code was actually written?

application/common/models/Reset.php
```diff
@@ -188,16 +186,59 @@ public function sendMethod()
...
+            'method' => 'phone',
+            'method_id' => $this->method_id,
+            'previous_attempts' => $this->attempts,
+        ];
+
+        // Get phone number without comma
+        $number = $this->method->getRawPhoneNumber();
+
+        // Generate random code for potential use
+        $code = Utils::getRandomDigits(\Yii::$app->phone->codeLength);
```
forevermatt: Should we add something to the PhoneVerificationInterface to make sure that every implementation will have a `codeLength` property?

application/common/models/Reset.php
```diff
@@ -188,16 +186,59 @@ public function sendMethod()
...
+
+        // Call component send() method to send verification and capture resulting code
+        $result = \Yii::$app->phone->send($number, $code);
+
+        // Update db with code and increased attempts count
+        $this->code = $result;
+        $this->attempts++;
```
forevermatt: Would it be worth updating the attempts count before we send the code to their phone number? As it currently stands, it looks like someone could attempt an unlimited number of times if there was some problem that was preventing the Reset from being updated in the database.

application/tests/mock/phone/Component.php
```diff
@@ -0,0 +1,62 @@
...
+     * Component may use $resetCode as a key for it's own purposes, as is the case with the Nexmo Verify service.
+     * Return true on success or throw NotMatchException when match fails.
+     * Throw \Exception when other exception occurs like network issue with service provider
+     * @param string $resetCode
+     * @param string $userProvided
+     * @return boolean
+     * @throws \Exception
+     * @throws \Sil\IdpPw\Common\PhoneVerification\NotMatchException
+     */
+    public function verify($resetCode, $userProvided)
```
forevermatt: In this mock function, why are we checking *both* whether the two provided values to match either other *and* whether they match the `id` and `code` values from any entries in our dummy test data file? It seems like this would (incorrectly) return true if the `$userProvided` value happened to match the `id` value in the dummy test data file.

application/common/models/Reset.php
fillup: I created github issues instead and removed @todo to clean up scrutinizer reports. I left a couple @todo in places where there was code but something needed to be finished and removed it from places where there was no code at all and would not be confused as done.

application/common/models/Reset.php
fillup: I dont think we technically can right? Other than docs.

application/common/models/Reset.php
fillup: I created an issue to make sure we review this and fix if needed

application/tests/mock/phone/Component.php
fillup: I was trying to cover both types of cases where the component may use the $resetCode as an internal ID to look up the reset code, like Nexmo Verify does, as well as just compare the two values like a basic SMS component may do.

application/common/models/Reset.php
forevermatt: I think you're right that we couldn't do it directly as an attribute, but we could add a `getCodeLength()` method.

application/tests/mock/phone/Component.php
forevermatt: What if we checked the `$resetCode` against the dummy `id` values first. If it matched an `id`, then we go ahead and return true or throw a `NotMatchException` based on whether `$userProvided` matches the value in that dummy data entry. If `$resetCode` didn't match any `id` values, then we could return/throw based on whether `$resetCode` matched `$userProvided`. That should at least avoid some false positives, though it still seems like there's probably some possibility for a wrong response if someone really tried to get one (which wouldn't make sense, since this is just a mock for tests).

fillup: thats fine, I'll swap the order

PR #18, opened 2016-04-20
email method support for resets

fillup:
* Added EmailQueue model and table
 * Updated Reset model to use EmailQueue to send emails for all email related methods
 * Fixed rebuildbasemodels.sh script so some other base models were updated
 * Added EventLog model and tests

@Baggerone @forevermatt @wcjr


application/common/config/main.php
```diff
@@ -1,16 +1,28 @@
...
-$adminEmail = getenv('ADMIN_EMAIL');
-$appEnv = getenv('APP_ENV');
-$recaptchaSiteKey = getenv('RECAPTCHA_SITE_KEY');
-$recaptchaSecretKey = getenv('RECAPTCHA_SECRET_KEY');
-$uiUrl = getenv('UI_URL');
+$mysqlHost = Env::get('MYSQL_HOST');
+$mysqlDatabase = Env::get('MYSQL_DATABASE');
+$mysqlUser = Env::get('MYSQL_USER');
+$mysqlPassword = Env::get('MYSQL_PASSWORD');
+$mailerUseFiles = Env::get('MAILER_USERFILES', false);
```
Baggerone: typo?  Use vs. USER

fillup: Yep, thanks.

application/common/mail/reset/self.php
```diff
@@ -0,0 +1,15 @@
...
+Someone recently requested a password change for your <?php echo \Yii::$app->params['appName']; ?>
+account. If this was you, you can use the reset code below to set a new password.
+
+Reset Code: <?php echo \yii\helpers\Html::encode($resetCode); ?> 
+
+If you don't want to change your password or didn't request this, just 
+ignore and delete this message. 
+
+To keep your account secure, please don't forward this email to anyone. 
+See our Help Center for <a href="<?php echo \Yii::$app->params['helpCenterUrl']; ?>">more security tips</a>.
```
Baggerone: Is it worth wrapping this in an if block, in case there is no helpCenterUrl

application/common/mail/reset/on-behalf.php
```diff
@@ -0,0 +1,15 @@
...
+Please contact them directly to ensure that you are only providing the following reset code to them
+and not to someone else.
+
+Reset Code: <?php echo \yii\helpers\Html::encode($resetCode); ?>
+
+To keep their account secure, please don't forward this email to anyone.
+See our Help Center for <a href="<?php echo \Yii::$app->params['helpCenterUrl']; ?>">more security tips</a>.
+
+Thanks!
+- <?php echo \yii\helpers\Html::encode(\Yii::$app->params['fromName']); ?>
```
Baggerone: it would be nice to have a shorter way of calling the `encode` function

application/common/models/EmailQueue.php
```diff
@@ -0,0 +1,219 @@
...
+        $emailQueue->text_body = $textBody;
+        $emailQueue->html_body = $htmlBody;
+        $emailQueue->cc_address = $ccAddress;
+        $emailQueue->event_log_user_id = $eventLogUserId;
+        $emailQueue->event_log_topic = $eventLogTopic;
+        $emailQueue->event_log_details = $eventLogDetails;
+
+        $emailQueue->send();
+
+        return $emailQueue;
```
Baggerone: This still needs the part that saves a new EmailQueue on send failure, right?

application/common/models/EmailQueue.php
```diff
@@ -0,0 +1,219 @@
...
+            'subject' => $this->subject,
+        ];
+
+        /*
+         * Try to send email or throw exception
+         */
+        try {
+            $mailer = $this->getMailer();
+            $sent = $mailer->send();
+            if ($sent === 0 || $sent === false) {
```
Baggerone: if the case where `$sent === None` is the same as these, you can just say `if ( ! $sent) {`

application/common/mail/reset/on-behalf.php
fillup: Technically we can "use yii\helpers\Html" and then just "Html::encode(" but there are some scrutinizer warnings about another class named Html already existing in the namespace so we just use it explicitly.

application/common/models/EmailQueue.php
```diff
@@ -0,0 +1,219 @@
...
+        return $mailer;
+    }
+
+    /**
+     * Creates an EventLog entry if $this->event_log_* properties are set
+     */
+    private function createEventLogEntry()
+    {
+        if ($this->event_log_user_id !== null && $this->event_log_topic !== null && $this->event_log_details !== null) {
+//            EventLog::log($this->event_log_user_id, $this->event_log_topic, $this->event_log_details);
```
Baggerone: Do you still want this commented out?

application/common/models/EmailQueue.php
fillup: hmm, I have the queue in send on a failure, I suppose it could be elevated to here if an exception occurs during send so that send is explicitly only sending, not sure if it matters though

application/common/models/EmailQueue.php
fillup: Yeah I was trying to be explicit because I saw that it can either return the number of emails sent or false on error, but it also seems to throw exceptions, so I was trying to be explicit and cover the bases.

application/common/mail/reset/self.php
fillup: I was just planning that there would be one as part of our default config. Technically it could not exist and then cause an error here, so perhaps I should, but it might be better to get an error so it can be fixed rather than silently working.

application/common/models/EmailQueue.php
```diff
@@ -0,0 +1,219 @@
...
+    {
+        if ( ! $this->save()) {
+            /*
+             * Queue failed, log it and throw exception
+             */
+            $log['status'] = 'failed to queue';
+            $log['error'] = Json::encode($this->getFirstErrors());
+            \Yii::error($log, 'application');
+            throw new \Exception('Unable to send or queue email', 1461009236);
+        } else {
```
Baggerone: Do you want me to stop pointing out unnecessary `else` blocks. :-)

application/common/models/EmailQueue.php
fillup: Yes, EventLog doesn't exist yet. I was going to do it with this PR but thought I would keep it separate.

application/common/models/EmailQueue.php
fillup: keep it up, I keep refactoring things so I miss these sometimes

application/common/mail/reset/on-behalf.php
Baggerone: could we say `use yii\helpers\Html as yHtml` and then `yHtml::encode`?

Baggerone:
I skimmed through the EventLog stuff and didn't notice anything worth commenting on.

application/common/mail/reset/on-behalf.php
```diff
@@ -0,0 +1,15 @@
+Hi there,
+
+<?php echo \yii\helpers\Html::encode($name); ?> recently requested a password change for their
+<?php echo \Yii::$app->params['appName']; ?> account and they have requested your assistance.
```
forevermatt: So would `appName` refer to the overall system being affected by these password resets, or does it refer to this password reset system itself?  I assumed `appName` would be something like "[System] Password Manager", not just "[System]", but the way it's used here seems to indicate otherwise.

application/common/mail/reset/self.php
forevermatt: It might be worth verifying what actually happens. It might not throw an error, but just insert an empty string there. The `\Yii::$app->params` object isn't quite a normal array, as far as I'm aware. I think it's some kind of collection class.

application/common/models/EmailQueue.php
```diff
@@ -0,0 +1,219 @@
...
+        }
+    }
+
+    /**
+     * Builds a mailer object from $this and returns it
+     * @return \yii\mail\MessageInterface
+     */
+    private function getMailer()
+    {
+        $mailer = \Yii::$app->mailer->compose();
```
forevermatt: For what it's worth, this actually creates a message ([`yii\mail\MessageInterface`](http://www.yiiframework.com/doc-2.0/yii-mail-messageinterface.html)), not a mailer (`yii\mail\MailerInterface`), so it may be worth renaming this from `$mailer` to `$message` (and possibly changing the name of `getMailer()` as well).

application/common/models/EmailQueue.php
forevermatt: Where are you seeing that it returns the number of emails sent? In [the documentation](http://www.yiiframework.com/doc-2.0/yii-mail-messageinterface.html#send%28%29-detail), all I see is a boolean return value.

application/common/models/EmailQueue.php
forevermatt: I think I agree with Steve on this one.  Having the `sendOrQueue` function actually handle the queuing of the message (if sending failed) feels like the right place for that code.  As an example of what might go wrong otherwise, it seems like we may run the risk of a queued email being requeued every time it fails to send (during normal processing of the email queue), resulting in a message having several entries in the email queue.

application/common/models/EmailQueue.php
```diff
@@ -0,0 +1,219 @@
...
+        
+        return $mailer;
+    }
+
+    /**
+     * Creates an EventLog entry if $this->event_log_* properties are set
+     */
+    private function createEventLogEntry()
+    {
+        if ($this->event_log_topic !== null && $this->event_log_details !== null && $this->event_log_user_id !== null) {
```
forevermatt: Should there be some indication to the code calling this if no event log entry ends up being created? I'm not sure, it just seems like a potential for silent failure.

application/common/models/EmailQueue.php
```diff
@@ -0,0 +1,219 @@
...
+    private function queue($log = [], \Exception $previousException = null)
+    {
+        if ( ! $this->save()) {
+            /*
+             * Queue failed, log it and throw exception
+             */
+            $log['status'] = 'failed to queue';
+            $log['error'] = Json::encode($this->getFirstErrors());
+            \Yii::error($log, 'application');
+            throw new \Exception('Unable to send or queue email', 1461009236);
```
forevermatt: This error message should probably just be `'Unable to queue email'`, since that's all this function tries to do. If there is an error that prevents both sending and queuing an email, the function that was trying to do both (such as `sendOrQueue`) seems like the more appropriate place for that kind of error message to originate.

application/common/models/EmailQueue.php
```diff
@@ -0,0 +1,219 @@
+<?php
+namespace common\models;
+
+use common\helpers\Utils;
+use common\models\User;
+use yii\helpers\ArrayHelper;
+use yii\helpers\Json;
+
+class EmailQueue extends EmailQueueBase
```
forevermatt: Wacky suggestion, feel free to reject it: 
Would it make more sense to name these things `QueuedEmail`s? Overall, the table acts like an email queue, but that's not how table names end up being read. This isn't a table of email queues.

Our `User` table contains users, our `Method` table contains methods... so the table for queued emails would be `QueuedEmail`.  This just feels like it might make things more readable, because instance of models that represent rows in this table would now be named according to how we're actually using them.

Thoughts?  (Again, feel free to reject this or say we'll consider it some other time.)

application/common/models/Reset.php
```diff
@@ -99,7 +101,7 @@ public static function findOrCreate($user, $type = self::TYPE_PRIMARY, $methodId
              * If $method_id is provided, make sure user owns it
              */
             if ($type == self::TYPE_METHOD && $methodId !== null) {
-                $method = Method::findOne(['user_id' => $user->id, 'id' => $methodId]);
+                $method = Method::findOne(['user_id' => $user->id, 'id' => $methodId, 'verified' => 1]);
```
forevermatt: Good catch.

application/common/models/Reset.php
```diff
@@ -146,23 +148,49 @@ public function send()
...
     public function sendSpouse()
     {
-        /*
-         * check if user has a spouse and then email the spouse the reset code
-         */
+        if ($this->user->hasSpouse()) {
+            $spouse = $this->user->getSpouseEmail();
+            $this->sendOnBehalf($spouse);
+        } else {
+            throw new \Exception('User does not have supervisor on record', 1461173477);
```
forevermatt: Looks like `supervisor` here was meant to be `spouse`.

application/common/models/EmailQueue.php
fillup: I cant find it now, I'll simplify to just ```if ( ! $mailer->send()) {```

application/common/mail/reset/on-behalf.php
fillup: You're right, I already decided that needed to be renamed to make more sense, I'm, thinking something like ```accountName``` instead

application/common/models/EmailQueue.php
fillup: fixed

application/common/models/EmailQueue.php
fillup: Sure, I'll have it return false in an ```else```. If there is an error creating the entry it throws an exception.

application/common/models/EmailQueue.php
wcjr: I think that is much more clarifying, not wacky at all!

application/common/models/Reset.php
fillup: yep

application/common/models/EmailQueue.php
```diff
@@ -149,42 +183,47 @@ private function getMailer()
 
     /**
      * Creates an EventLog entry if $this->event_log_* properties are set
+     * @return boolean
+     * @throws \Exception
      */
     private function createEventLogEntry()
     {
         if ($this->event_log_topic !== null && $this->event_log_details !== null && $this->event_log_user_id !== null) {
             EventLog::log($this->event_log_topic, $this->event_log_details, $this->event_log_user_id);
```
forevermatt: You might want to add a `return true;` after this line, now that there's a `return false;` when no `EventLog` entry was written.

application/common/models/Reset.php
```diff
@@ -176,19 +204,64 @@ public function sendMethod()
...
         /*
-         * send email to $this->method->value with reset code
+         * Generate code if needed, update attempt counter, save record, and send email
          */
+        if ($this->code === null) {
+            $this->code = Utils::getRandomDigits(\Yii::$app->params['reset']['codeLength']);
+        }
+        $this->attempts += 1;
+        if ($this->save()) {
+            $body = \Yii::$app->mailer->render(
```
forevermatt: Is this generating text or HTML email content? It seems odd to me that `$body` is used below as both the textBody and the htmlBody parameters to `EmailQueue::sendOrQueue(...)`, thus the question.

application/console/migrations/m160414_140140_email_queue.php
```diff
@@ -0,0 +1,37 @@
...
+        $this->createTable(
+            '{{email_queue}}',
+            [
+                'id' => 'pk',
+                'to_address' => 'varchar(255) not null',
+                'cc_address' => 'varchar(255) null',
+                'subject' => 'varchar(255) not null',
+                'text_body' => 'text null',
+                'html_body' => 'text null',
+                'attempts_count' => 'tinyint not null default 0',
```
forevermatt: Is there any noticeable risk that someone trying to hack the system could get the `attempts_count` to overflow the range of a tinyint? I just know that's one of the examples used in textbooks about how an error counter can exceed it's data type size and get back to a state making it seem like there were no errors (or in our case, no attempts). Probably a pretty unlikely situation, but the use of tinyint (0 - 255) brought that to mind.

application/tests/mock/phone/Component.php
```diff
@@ -57,6 +52,11 @@ public function verify($resetCode, $userProvided)
             }
```
forevermatt: I don't know how much it matters, but I was also suggesting checking `$phone['id'] == $resetCode` here first (separately), and if it matched then either returning or throwing (based on `$phone['code'] == $userProvided`), rather than just continuing to check for possible matches. So if a matching `id` was found, the function was going to exit one way or another. Otherwise we still have the same "vulnerability" (except that it's probably not significant enough to be worth calling it that).

application/tests/unit/common/models/EmailQueueTest.php
```diff
@@ -0,0 +1,178 @@
...
+            'text_body' => 'test body',
+        ];
+
+        EmailQueue::sendOrQueue(
+            $data['toAddress'],
+            $data['subject'],
+            $data['text_body']
+        );
+
+        $this->assertTrue(EmailUtils::hasEmailFileBeenCreated($data['subject']));
```
forevermatt: If you really wanted to be thorough, you could do this same assertion (except as an `assertFalse`) just before the call to `EmailQueue::sendOrQueue(...)` to make sure the system is in the expected state before you call the function.

application/tests/unit/common/models/EventLogTest.php
```diff
@@ -0,0 +1,63 @@
...
+         */
+        $eventLog = EventLog::findOne(['topic' => 'test-1461183312']);
+
+        $this->assertNotNull($eventLog);
+    }
+
+    public function testLogInvalidUser()
+    {
+        $this->setExpectedException('\Exception', '', 1461182172);
+        EventLog::log('test-1461183456', 'testing-1461183456', 44);
```
forevermatt: If our list of user fixtures ever reaches the point of having 44 users in it, would this test then fail?

common.env
```diff
@@ -4,3 +4,4 @@ MYSQL_DATABASE=idpmgmt
 MYSQL_USER=idpmgmt
 MYSQL_PASSWORD=idpmgmt
 MAILER_USEFILES=true
+APP_NAME=SIL
```
forevermatt: Does this need to be changed to `IDP_NAME`?

forevermatt:
@fillup Okay, I'm done looking through this. Once you think the comments are sufficiently addressed, feel free to merge. =o)

application/console/migrations/m160414_140140_email_queue.php
fillup: I think the plan is to have a limit and check what the counter is before trying again, and once hit limit stop trying. Those checks and such still needed to be added though and we can talk through the logic of exponential back off on retries and what the limit should be.

application/tests/mock/phone/Component.php
fillup: This is just for unit tests, so I think it is sufficient as is.

common.env
fillup: yea, I updated in wrong file.

application/tests/unit/common/models/EventLogTest.php
fillup: yeah, I changed to 4499999999

application/tests/unit/common/models/EmailQueueTest.php
fillup: true

application/common/models/Reset.php
fillup: good point, I'll create an issue to revisit if we need to render different versions

PR #22, opened 2016-05-04
Method APIs built out

fillup:
Ability to add, verify, and delete methods via API. Works with email, Nexmo Verify, and Nexmo SMS.

@Baggerone @wcjr @forevermatt


application/common/models/Method.php
```diff
@@ -61,4 +93,156 @@ public function getRawPhoneNumber()
...
+        $method->user_id = $userId;
+        $method->type = $type;
+        $method->value = $value;
+
+        if ($type == self::TYPE_PHONE) {
+            $log['value'] = Utils::maskPhone($value);
+        } elseif ($type == self::TYPE_EMAIL) {
+            $log['value'] = Utils::maskEmail($value);
+        } else {
+            $log['value'] = 'invalid type';
```
Baggerone: Were these just for logging an error, or did you forget to add code to log success?

application/common/models/Method.php
```diff
@@ -61,4 +93,156 @@ public function getRawPhoneNumber()
...
+     * Delete all method records that are not verified and verification_expires date is in the past
+     * @throws \Exception
+     */
+    public static function deleteExpiredUnverifiedMethods()
+    {
+        $methods = Method::find()->where(['verified' => 0])
+                                 ->andWhere(['<', 'verification_expires', Utils::getDatetime()])
+                                 ->all();
+
+        foreach ($methods as $method) {
```
Baggerone: I vaguely remember that yii gives you a delete() method that means you don't have to have this separate foreach loop.

application/frontend/controllers/MethodController.php
```diff
@@ -38,6 +42,159 @@ public function actionIndex()
...
+        } catch (\Exception $e) {
+            \Yii::error([
+                'action' => 'deleteExpiredUnverifiedMethods',
+                'status' => 'error',
+                'error' => $e->getMessage(),
+            ]);
+        }
+
+        /*
+         * Find method if belongs to user and is not expired
```
Baggerone: Does the comment need correcting? I don't see where expired comes into this.

application/frontend/controllers/MethodController.php
```diff
@@ -38,6 +42,159 @@ public function actionIndex()
...
+        $method = Method::findOne([
+            'uid' => $uid,
+            'user_id' => \Yii::$app->user->getId(),
+        ]);
+
+        /*
+         * If not found, throw 404
+         */
+        if ($method === null) {
+            throw new NotFoundHttpException();
```
Baggerone: Do you want to throw the same exception for unfound method and invalid code, so that a hacker doesn't know when he has a good method.

application/frontend/controllers/MethodController.php
```diff
@@ -38,6 +42,159 @@ public function actionIndex()
...
+     * @return array
+     * @throws NotFoundHttpException
+     * @throws \Exception
+     */
+    public function actionDelete($uid)
+    {
+        /** @var Method $method */
+        $method = Method::findOne([
+            'uid' => $uid,
+            'user_id' => \Yii::$app->user->getId()
```
Baggerone: Is it necessary to get a match on the user_id?

application/tests/unit/common/models/MethodTest.php
```diff
@@ -57,5 +60,131 @@ public function testGetMaskedValueEmail()
...
+        $this->assertNotNull($method->verification_expires);
+
+        $this->setExpectedException('\Exception', '', 1461442988);
+        $method->validateAndSetAsVerified('asdf1234');
+        $this->assertEquals(2, $method->verification_attempts);
+    }
+
+    public function testDeleteExpiredUnverifiedMethods()
+    {
+        $method1 = Method::findOne(['uid' => '33333333333333333333333333333333']);
```
Baggerone: you could assign the '333...' to a variable, so a code reviewer doesn't need to count to see if they're the same. :-)

dockerbuild/logentries.all.crt
```diff
@@ -0,0 +1,91 @@
...
+ADCCAQoCggEBAI7CAhnhoFmk6zg1jSz9AdDTScBkxwtiBUUWOqigwAwCfx3M28Sh
+bXcDow+G+eMGnD4LgYqbSRutA776S9uMIO3Vzl5ljj4Nr0zCsLdFXlIvNN5IJGS0
+Qa4Al/e+Z96e0HqnU4A7fK31llVvl0cKfIWLIpeNs4TgllfQcBhglo/uLQeTnaG6
+ytHNe+nEKpooIZFNb5JPJaXyejXdJtxGpdCsWTWM/06RQ1A/WZMebFEh7lgUq/51
+UHg+TLAchhP6a5i84DuUHoVS3AOTJBhuyydRReZw3iVDpA3hSqXttn7IzW3uLh0n
+c13cRTCAquOyQQuvvUSH2rnlG51/ruWFgqUCAwEAAaOCAWUwggFhMB8GA1UdIwQY
+MBaAFLuvfgI9+qbxPISOre44mOzZMjLUMB0GA1UdDgQWBBSQr2o6lFoL2JDqElZz
+30O0Oija5zAOBgNVHQ8BAf8EBAMCAYYwEgYDVR0TAQH/BAgwBgEB/wIBADAdBgNV
+HSUEFjAUBggrBgEFBQcDAQYIKwYBBQUHAwIwGwYDVR0gBBQwEjAGBgRVHSAAMAgG
+BmeBDAECATBMBgNVHR8ERTBDMEGgP6A9hjtodHRwOi8vY3JsLmNvbW9kb2NhLmNv
```
Baggerone: I think there's a typo in here. (-;

application/common/models/Method.php
fillup: good catch, I forgot it

application/common/models/Method.php
fillup: True, but I specifically want to use the model's delete method so it can do anything else needed like deleting foreign key records and such.

application/frontend/controllers/MethodController.php
fillup: These API calls require you to be logged in, so I dont think that is as much of a concern. We'll certainly do that with Resets though.

application/frontend/controllers/MethodController.php
fillup: Yes, only want the user to delete their own methods.

application/tests/unit/common/models/MethodTest.php
fillup: yeah good call

dockerbuild/logentries.all.crt
fillup: Spellchecker says its good, what language do you have yours configured for?

application/common/helpers/Utils.php
```diff
@@ -138,9 +139,15 @@ public static function maskPhone($phone)
     /**
      * @param string $email an email address (it doesn't verify it)
```
forevermatt: This comment is no longer correct (since it does verify... or validate?... the email address).

application/common/mail/method/verify.php
```diff
@@ -0,0 +1,15 @@
...
+Someone recently requested to add this email address, <?php echo \yii\helpers\Html::encode($toAddress); ?>,
+as a method for verifying themselves should they need to reset their <?php echo \Yii::$app->params['idpName']; ?>
+account. If this was you, you can use the reset code below to set a new password.
+
+Reset Code: <?php echo \yii\helpers\Html::encode($code); ?>
+
+If you did not request adding this email address to your account please delete this email.
+
+To keep your account secure, please don't forward this email to anyone.
+See our Help Center for <a href="<?php echo \Yii::$app->params['helpCenterUrl']; ?>">more security tips</a>.
```
forevermatt: The helpCenterUrl value might also need to be escaped, so that any special characters (such as ampersands) are properly encoded for use in an HTML attribute.

Just from a separation of concerns standpoint, it would probably also be better if the helpCenterUrl was passed into this view rather than having this view reach out for it. We recently had to fix a bug in another project that was partially caused by a view file reaching out to somewhere else in the code rather than being given all the data that it needed.

application/common/mail/method/verify.php
```diff
@@ -0,0 +1,15 @@
...
+
+Thanks!
+- <?php echo \yii\helpers\Html::encode(\Yii::$app->params['fromName']); ?>
```
forevermatt: Similar to above, it may be better to pass the fromName into the view rather than having the view file reach out for it.

application/common/models/Method.php
```diff
@@ -25,10 +25,33 @@ public function rules()
                 ],
 
                 [
+                    ['verified', 'verification_attempts'], 'default', 'value' => 0,
+                ],
+
+                [
+                    'verification_code', 'default', 'when' => function() { return $this->getIsNewRecord(); },
```
forevermatt: Does Yii2 to require you to use this `'when' => function() { return $this->getIsNewRecord(); }` syntax rather than some sort of `'on' => 'insert'` syntax the way Yii1 worked?

application/common/models/Method.php
```diff
@@ -61,4 +93,159 @@ public function getRawPhoneNumber()
...
+     * @param string $type
+     * @param string $value
+     * @return Method
+     * @throws \Exception
+     */
+    public static function createAndSendVerification($userId, $type, $value)
+    {
+        $log = [
+            'class' => __CLASS__,
+            'method' => 'createAndSendVerification',
```
forevermatt: I think you can use `__METHOD__` here if you don't want to retype the method name for these sorts of log entries. (http://php.net/manual/en/language.constants.predefined.php)

application/common/models/Method.php
```diff
@@ -61,4 +93,159 @@ public function getRawPhoneNumber()
...
+    /**
+     * Validate user submitted code and update record to be verified if valid
+     * @param string $userSubmitted
+     * @throws \Exception
+     */
+    public function validateAndSetAsVerified($userSubmitted)
+    {
+        if ( ! $this->isUserProvidedCodeCorrect($userSubmitted)) {
+            $this->verification_attempts++;
+            $this->save();
```
forevermatt: Do we need to check the return value of `save()` here?

application/common/models/Method.php
```diff
@@ -61,4 +93,159 @@ public function getRawPhoneNumber()
...
+     * @throws \Exception
+     */
+    public static function deleteExpiredUnverifiedMethods()
+    {
+        $methods = self::find()->where(['verified' => 0])
+                               ->andWhere(['<', 'verification_expires', Utils::getDatetime()])
+                               ->all();
+
+        foreach ($methods as $method) {
+            $method->delete();
```
forevermatt: Do we need to do something (or know about it) if the `delete()` fails?

application/common/models/Reset.php
```diff
@@ -310,7 +312,9 @@ public function sendPhone()
     public function isUserProvidedCodeCorrect($userProvided)
     {
         if ($this->type == self::TYPE_METHOD && $this->method->type == Method::TYPE_PHONE) {
-            return \Yii::$app->phone->verify($this->code, $userProvided);
+            return Verification::isPhoneCodeValid($this->code, $userProvided);
+        } else {
```
forevermatt: Do we need to make separately handle email methods and unknown methods here like we did in `Method->sendVerification()` above?

application/common/models/Verification.php
```diff
@@ -0,0 +1,113 @@
...
+                'code' => $code,
+                'toAddress' => $toAddress,
+            ]
+        );
+
+        EmailQueue::sendOrQueue(
+            $toAddress,
+            $subject,
+            $body,
+            $body,
```
forevermatt: I'm guessing it doesn't matter, but do you know if it will cause any problems if the textBody of an email contains HTML? Some of our views contain link tags, and we're using the same body for htmlBody and textBody. Just a thought.

application/common/models/Verification.php
```diff
@@ -0,0 +1,113 @@
...
+    public static function sendPhone(
+        $phoneNumber,
+        $code,
+        $eventLogUserId = null,
+        $eventLogTopic = null,
+        $eventLogDetails = null
+    ) {
+        $result = \Yii::$app->phone->send($phoneNumber, $code);
+
+        if ($eventLogUserId !== null && $eventLogTopic !== null && $eventLogDetails !== null) {
```
forevermatt: Would we still want to log an EventLog entry even if no `$eventLogDetails` were provided, but the other two values were?

application/frontend/controllers/MethodController.php
```diff
@@ -38,6 +40,159 @@ public function actionIndex()
...
+                sprintf('Type is required. Options are: %s or %s', Method::TYPE_EMAIL, Method::TYPE_PHONE)
+            );
+        } elseif ($value === null) {
+            throw new BadRequestHttpException('Value is required.');
+        }
+
+        /*
+         * Check for existing method with this value
+         */
+        $method = Method::findOne(['value' => $value, 'user_id' => \Yii::$app->user->getId()]);
```
forevermatt: Do we care whether it the `$type` also matches?

application/frontend/controllers/MethodController.php
```diff
@@ -38,6 +40,159 @@ public function actionIndex()
...
+        $request = \Yii::$app->request;
+        $code = $request->getBodyParam('code');
+        if ($code === null) {
+            throw new BadRequestHttpException('Code is required');
+        }
+
+        try {
+            $method->validateAndSetAsVerified($code);
+        } catch (\Exception $e) {
+            throw new BadRequestHttpException('Invalid verification code');
```
forevermatt: An exception could also be thrown by `validateAndSetAsVerified(...)` (leading to this code here being run) if there was a problem saving changes to the method. Would we still want to use the error message defined here (`'Invalid verification code'`) in those situations?

application/frontend/controllers/MethodController.php
```diff
@@ -38,6 +40,159 @@ public function actionIndex()
...
+        $method = Method::findOne([
+            'uid' => $uid,
+            'user_id' => \Yii::$app->user->getId()
+        ]);
+
+        if ($method === null) {
+            throw new NotFoundHttpException();
+        }
+
+        $method->delete();
```
forevermatt: Do we need to check the return value of `delete()`?

application/tests/unit/common/models/MethodTest.php
```diff
@@ -57,5 +60,133 @@ public function testGetMaskedValueEmail()
...
+        $user = $this->users('user1');
+        $mockPhones = include __DIR__ . '/../../../mock/phone/data.php';
+        $method = Method::createAndSendVerification(
+            $user->id,
+            Method::TYPE_PHONE,
+            $mockPhones[0]['number']
+        );
+
+        $this->assertEquals($mockPhones[0]['code'], $method->verification_code);
+    }
```
forevermatt: I may just be missing it, but what is it about the `createAndSendVerification` function that this test is proving works as intended?

application/tests/unit/common/models/MethodTest.php
```diff
@@ -57,5 +60,133 @@ public function testGetMaskedValueEmail()
...
+    {
+        $user = $this->users('user1');
+        $mockPhones = include __DIR__ . '/../../../mock/phone/data.php';
+        $method = Method::createAndSendVerification(
+            $user->id,
+            Method::TYPE_PHONE,
+            $mockPhones[0]['number']
+        );
+
+        $this->assertEquals($mockPhones[0]['code'], $method->verification_code);
```
forevermatt: Forgive my ignorance here, but, similar to above (I think), why would the `verification_code` of the `Method` returned by `Method::createAndSendVerification(...)` match the `'code'` in the mock phone data?

forevermatt:
@fillup Okay. Once you consider all my comments sufficiently addressed, I'm fine with you merging this.

application/common/helpers/Utils.php
fillup: fixed

application/common/mail/method/verify.php
fillup: Sure, fixed.

application/common/mail/method/verify.php
fillup: Also fixed.

application/common/models/Method.php
fillup: I believe in Yii2 the 'on' attribute must reference a defined scenario: http://www.yiiframework.com/doc-2.0/guide-structure-models.html#scenarios

So I don't think we can just do on insert or on update.

application/common/models/Method.php
fillup: We're already throwing an exception right after it so I'm not sure what we'd do if it failed anyway, ideas?

application/common/models/Method.php
fillup: sure, I'll log it, but I dont want to throw an exception and prevent user from moving forward

application/common/models/Reset.php
fillup: i suppose

application/common/models/Verification.php
fillup: My expectation is it just would display all the html as text if their client only supported text emails, which I'm okay with for now because I'm not aware of any clients like that that would be in use by our folks.

application/common/models/Verification.php
fillup: Details are a required field in the db. If we think we'd like to just log that an event occurred without any details we'd have to modify the db and EventLog model too.

application/frontend/controllers/MethodController.php
fillup: I dont think that would matter since an email address and phone number cannot be equal, it would not get confused here.

application/frontend/controllers/MethodController.php
fillup: Sure. I added a new exception type of InvalidCodeException that I'll catch for first and provide different errors.

application/frontend/controllers/MethodController.php
fillup: Yes, I wish the model methods just threw exceptions themselves. I added a check and exception for failure.

application/tests/unit/common/models/MethodTest.php
fillup: It is the mock phone verification class that is assigning a verification code, so this test is just verifying that it does.

application/common/models/Method.php
forevermatt: If the save failed, the user could keep trying codes ad infinitum until they got one right. One option would be to update the `verification_attempts` and save that change before even checking the `$userSubmitted` code. That way, if we failed to update our attempts counter, it would be impossible for them to validate the method. It's that whole practice of "fail closed" (meaning if something goes wrong, default to blocking the user instead of letting them proceed without actually authenticating them).

application/tests/unit/common/models/MethodTest.php
forevermatt: Okay. Should tests of our mock classes be put somewhere separate from tests of our Method model? Or did I misunderstand your answer?

application/common/models/Method.php
fillup: Sure I'll fix it. This particular method/api requires that the user already be logged in too, so it doesn't seem like much of a risk, but I'll do it here since I'll do the same with resets which are the bigger concern. I'm pushing the update directly to develop since this came after branch was merged and deleted.

PR #25, opened 2016-05-11
Reset APIs

fillup:
Added support for creating reset, updating reset type/method, resending a reset, and validating a reset. 

@forevermatt @wcjr


application/frontend/controllers/ResetController.php
```diff
@@ -81,9 +127,77 @@ public function actionResend($uid)
...
+            throw new BadRequestHttpException('Code is required', 1462989866);
+        }
+
+        $log = [
+            'action' => 'Validate reset',
+            'reset_id' => $reset->id,
+        ];
+
+        $isValid = $reset->isUserProvidedCodeCorrect($code);
+        if ($isValid === true) {
```
wcjr: not sure this $isValid is adding much to readability here, I think you would be fine with

`if ($reset->isUserProvidedCodeCorrect($code))`

if you wanted to write a little less code.

fillup: I wanted to have an explicit check and not a loose check. Since that determines whether what they provided is correct or not, I want to be sure it is === true and not something considered truthy. I could do all that in the if call without variable assignment, but I thought the assignment made it more readable.

application/common/helpers/Utils.php
```diff
@@ -278,4 +283,73 @@ public static function getRandomDigits($length = 4)
...
+     * @param  string  $ip
+     * @return boolean
+     */
+    public static function isValidIpAddress($ip)
+    {
+        $flags = FILTER_FLAG_IPV4 | FILTER_FLAG_IPV6;
+        if (filter_var($ip, FILTER_VALIDATE_IP, $flags) === false) {
+            return false;
+        }
+        return true;
```
forevermatt: If you wanted, you could reduce these few lines to...
`return (filter_var($ip, FILTER_VALIDATE_IP, $flags) !== false);`

I'm not sure if that's more or less readable, though, which seems more important than conciseness.

application/common/models/Reset.php
```diff
@@ -2,10 +2,13 @@
 namespace common\models;
 
 use common\helpers\Utils;
+use GuzzleHttp\Tests\Ring\Client\Server;
```
forevermatt: I don't see anywhere that this `use` statement's class is actually being used. Am I just missing it?

application/common/models/Reset.php
```diff
@@ -229,40 +250,37 @@ public function sendMethod()
...
-                $this->user->id,
-                self::TOPIC_RESET_EMAIL_SENT,
-                'Password reset email for ' . $this->user->getDisplayName() .
-                'sent to ' . $toAddress
-            );
 
-        } else {
-            throw new \Exception('Unable to update reset in database, email not sent', 1461098651);
-        }
+        $this->saveOrError('send email','Unable to update reset in database, email not sent.');
```
forevermatt: If the code wasn't `null`, is there still a reason to save the Reset here?

application/common/models/Reset.php
fillup: Good catch, probably an IDE auto completion.

application/common/models/Reset.php
```diff
@@ -229,40 +250,37 @@ public function sendMethod()
...
+            $toAddress,
+            $subject,
+            '@common/mail/reset/' . $view,
+            $this->code,
+            $this->user,
+            $ccAddress,
+            $this->user->id,
+            self::TOPIC_RESET_EMAIL_SENT,
+            'Password reset email for ' . $this->user->getDisplayName() .
+            'sent to ' . $toAddress
```
forevermatt: Looks like you may need another space here: `' sent`.
Similarly below in `sendPhone()`.

application/common/helpers/Utils.php
fillup: sure.

application/common/models/Reset.php
```diff
@@ -290,14 +308,8 @@ public function sendPhone()
 
         // Update db with code and increased attempts count
```
forevermatt: The second half of this comment is now being done before this function is called.

application/common/models/Reset.php
fillup: I suppose not

application/common/models/Reset.php
```diff
@@ -337,4 +377,156 @@ public static function getExpireTimestamp()
...
+     */
+    public function isDisabled()
+    {
+        if ($this->disable_until !== null) {
+            $disableUntilTime = strtotime($this->disable_until);
+            // Intentionally loose comparison to catch zero
+            if ($disableUntilTime == false) {
+                return true;
+            }
+            return $disableUntilTime > time();
```
forevermatt: One thing this logic assumes is that the timezone used for storing `disable_until` in the database is the same timezone as used by `time()`. That may be a safe assumption... but I'm not sure of that. Is there some explicit way we could be using UTC for all of these time values?

application/common/models/Reset.php
fillup: thanks

application/common/models/Reset.php
fillup: thanks

application/common/models/Reset.php
fillup: We generate the datetime entry from php and then insert it, so we don't use the db to generate dates/times at all.

application/common/models/Reset.php
```diff
@@ -337,4 +375,208 @@ public static function getExpireTimestamp()
...
+                'error' => 'Reset is currently disabled until ' . $this->disable_until,
+            ]);
+            throw new TooManyRequestsHttpException();
+        }
+    }
+
+    /**
+     * Save model or throw exception on error
+     * @param string $action
+     * @param string $errorPrefix
```
forevermatt: Do you know if the `message` property of a `ServerErrorHttpException` is returned in the HTTP Response? If so, we might want to add a comment here in the doc. block that the value of `$errorPrefix` will probably be displayed to the user (just as a reminder to avoid putting anything sensitive in the value passed in for `$errorPrefix`). Just a thought.

application/common/models/User.php
```diff
@@ -168,7 +168,18 @@ public function updateProfileIfNeeded($firstName, $lastName, $username, $email)
      */
     public function getMaskedMethods()
     {
-        $methods = [];
+        /*
+         * Include primary email address
```
forevermatt: Do we also need to manually include the supervisor and spouse options (when applicable) here?

application/frontend/controllers/ResetController.php
```diff
@@ -81,9 +128,98 @@ public function actionResend($uid)
...
+                    'type' => $reset->type,
+                    'attempts' => $reset->attempts,
+                ],
+                $reset->user_id
+            );
+
+            /*
+             * Reset verified successfully, log user in
+             */
+            if (\Yii::$app->user->login($reset->user, \Yii::$app->params['sessionDuration'])) {
```
forevermatt: I'm not sure the second parameter to `yii\web\User->login(...)` actually affects the session lifetime. Per [the documentation](http://www.yiiframework.com/doc-2.0/yii-web-user.html#login%28%29-detail), it appears to simply add the ability to use a cookie to maintain the user's logged-in status, which I don't think is what we want.

It looks like `yii\web\User->authTimeout` is where we should be putting the value we're currently putting into our `params` under `'sessionDuration'`.

application/frontend/controllers/ResetController.php
```diff
@@ -81,9 +128,98 @@ public function actionResend($uid)
...
+                if ( ! $reset->delete()) {
+                    \Yii::error([
+                        'action' => 'delete reset after validation',
+                        'reset_id' => $reset->id,
+                        'status' => 'error',
+                        'error' => Json::encode($reset->getFirstErrors()),
+                    ]);
+                }
+
+                return new \stdClass();
```
forevermatt: Why return a `stdClass` here? It doesn't seem clear to me what `actionValidate(...)` should return (if anything). The doc. comment says a `Reset` instance will be returned, but that doesn't quite seem feasible to always do.

forevermatt:
Okay @fillup. I've commented on all of the issues that I see.

forevermatt:
We should also add sanity checks where applicable (such as the `send()` and `verify()` methods of any classes implementing the `PhoneVerificationInterface`) to make sure that `$code` isn't an empty string.

application/common/models/Reset.php
fillup: ok

application/common/models/User.php
fillup: I think best to just leave them out. We return boolean hasSupervisor and hasSpouse so the frontend can just tell us "supervisor" or "spouse", but I'm think it best not to display masked versions of the values. I might be convinced otherwise given a good use case, but for now I'll keep it limited.

application/frontend/controllers/ResetController.php
fillup: okay good catch, I updated both login places and the user component config for authTimeout

application/frontend/controllers/ResetController.php
fillup: fixed the phpdoc. I return a stdClass so it returns an empty object and 200 status

PR #26, opened 2016-05-20
Password APIs

fillup:
Adding password APIs for getting metadata and changing password. At this time the actual changing of the password is just stubbed out until the LDAP backend integration is ready.

@wcjr @forevermatt @Baggerone


application/common/models/Password.php
```diff
@@ -0,0 +1,155 @@
...
+
+    /** @var \Sil\IdpPw\Common\PasswordStore\PasswordStoreInterface */
+    public $passwordStore;
+
+    /** @var array */
+    public $config;
+
+    public function init()
+    {
+        //$this->passwordStore = \Yii::$app->passwordStore;
```
Baggerone: Is this needed?

Dockerfile
```diff
@@ -13,6 +13,10 @@ COPY dockerbuild/rsyslog.conf /etc/rsyslog.conf
 RUN mkdir -p /opt/ssl
 COPY dockerbuild/logentries.all.crt /opt/ssl/logentries.all.crt
 
+# get s3-expand
+RUN curl https://raw.githubusercontent.com/silinternational/s3-expand/master/s3-expand -o /usr/local/bin/s3-expand
```
forevermatt: Would it be worth pointing at a specific version of s3-expand, in case we even need to make a backwards-incompatible change to s3-expand?

application/common/config/local.test.php
```diff
@@ -1,6 +1,43 @@
...
+            'maxLength' => [
+                'value' => 255,
+                'phpRegex' => '',
+                'jsRegex' => '.{0,255}',
+                'enabled' => true
+            ],
+            'minNum' => [
+                'value' => 2,
+                'phpRegex' => '',
+                'jsRegex' => '(\d.*){2,}',
```
forevermatt: Have we tested this regex? I don't think it's actually enforcing what we want it to.

application/tests/unit/common/models/PasswordTest.php
```diff
@@ -0,0 +1,124 @@
...
+                'zxcvbn' => 4,
+                'minLength' => true,
+                'maxLength' => true,
+                'minNum' => true,
+                'minUpper' => true,
+                'minSpecial' => true,
+                'overall' => true,
+            ],
+            [
+                'password' => 'jesus',
```
Baggerone: Are you sure you want these references in here?

application/common/config/local.test.php
forevermatt: Hmm... well, when I try it out, it seems to work properly.

application/common/config/local.test.php
```diff
@@ -1,6 +1,43 @@
...
+                'enabled' => true
+            ],
+            'minUpper' => [
+                'value' => 1,
+                'phpRegex' => '',
+                'jsRegex' => '([A-Z].*){0,0}',
```
forevermatt: Should this have been `'jsRegex' => '([A-Z].*){1,}',` instead?

application/tests/unit/common/models/PasswordTest.php
```diff
@@ -0,0 +1,124 @@
...
+    public function testZxcvbn()
+    {
+        $this->markTestSkipped('Skipped for now');
+
+        $testData = $this->getTestData();
+
+        foreach($testData as $testCase) {
+            $zxcvbn = new Zxcvbn();
+            $strength = $zxcvbn->passwordStrength($testCase['password']);
+//            if ($testCase['password'] == '1John 3:16') {
```
Baggerone: Do you still want this?

application/common/config/local.test.php
```diff
@@ -1,6 +1,43 @@
...
+                'enabled' => true
+            ],
+            'minSpecial' => [
+                'value' => 1,
+                'phpRegex' => '',
+                'jsRegex' => '([\W_].*){0,0}',
```
forevermatt: Like above, it looks like `{0,0}` should be `{1,}` here (I think).

application/common/helpers/Utils.php
```diff
@@ -26,13 +26,14 @@ public static function getDatetime($timestamp = null)
     }
 
     /**
-     * @param integer|null $timestamp
+     * @param integer|string|null $timestamp time as unix timestamp, mysql datetime, or null for now
      * @return string
      */
     public static function getIso8601($timestamp = null)
     {
         $timestamp = $timestamp ?: time();
```
forevermatt: An unlikely edge case, but this would make it impossible to get the ISO 8601 value for a unix timestamp of zero (0). It might be better to just go ahead and actually check `$timestamp !== null`.

application/common/config/local.test.php
```diff
@@ -1,6 +1,43 @@
 <?php
 
 return [
+    'params' => [
+        'password' => [
+            'minLength' => [
+                'value' => 10,
+                'phpRegex' => '',
```
forevermatt: Since the Password model's rules will be trying to use these, should we go ahead and include the phpRegex values?

application/common/models/Password.php
```diff
@@ -0,0 +1,155 @@
...
+            [
+                'password',
+                ZxcvbnPasswordValidator::className(),
+                'minScore' => $this->config['zxcvbn']['minScore'],
+                'message' => \Yii::t(
+                    'app',
+                    'Your password does not meet the minimum strength of {minScore}',
+                    ['minScore' => $this->config['zxcvbn']['minScore']]
+                ),
+            ],
```
forevermatt: This entry looks like it's missing a `'when' => function() { return $this->config['zxcvbn']['enabled']; }`, though I don't know if we intend to allow this to be disabled.

application/common/models/Password.php
```diff
@@ -0,0 +1,155 @@
...
+        $password = new Password();
+        $password->password = $newPassword;
+
+        return $password;
+    }
+
+    /**
+     * @param int $userId
+     * @param string $username
+     * @return array
```
forevermatt: Why does this return an array?

application/common/models/Password.php
```diff
@@ -0,0 +1,155 @@
...
+     * @throws BadRequestHttpException
+     * @throws ServerErrorHttpException
+     */
+    public function save($userId, $username)
```
forevermatt: This function doesn't seem to do any validation to make sure that the given $userId and $username actually refer to the same user. That seems like something we should verify (if possible).

application/common/models/User.php
```diff
@@ -59,6 +59,9 @@ public function fields()
             'last_name',
             'idp_username',
             'email',
+            'password' => function ($model) {
```
forevermatt: Would it be good to call this `passwordMeta` instead of `password` (to avoid using the name "password" to refer to yet another thing)?

application/common/models/Password.php
fillup: Yes, it will be when I finish the backend integration.

Dockerfile
fillup: Sure, updated.

application/common/config/local.test.php
fillup: I think so, but since that is in the test file and for js I think it is just a copy/paste artifact.

application/common/config/local.test.php
fillup: yep

application/common/helpers/Utils.php
fillup: ok, I'll update

application/common/models/Password.php
fillup: Yeah we intended to not have it optional, but after some testing and finding that the php implementation does not calculate exactly the same as js version we may not be able to use it. We're considering running a separate nodejs container on backend with simple api and just use it to ensure we're using the exact same code on both front and back.

application/common/models/Password.php
fillup: no good reason I can think of, I'll remove it

application/common/models/Password.php
fillup: true, that function is only really stubbed out right now so it needs work.

application/common/models/User.php
fillup: Yeah probably. @wcjr you good with that change?

wcjr: fine with me, `password_meta` to be consistent?

application/common/config/local.test.php
fillup: yep, added.

application/common/models/User.php
fillup: ok, updated.

application/frontend/web/index.php
```diff
@@ -1,5 +1,5 @@
 <?php
-defined('YII_DEBUG') || define('YII_DEBUG', true);
+defined('YII_DEBUG') || define('YII_DEBUG', false);
```
wcjr: hmmm, have you tested this?

application/common/models/Password.php
```diff
@@ -0,0 +1,152 @@
...
+                'message' => \Yii::t(
+                    'app',
+                    'Your password must contain at least {minSpecial} special characters (code 140)',
+                    ['minSpecial' => $this->config['minSpecial']['value']]
+                ),
+                'when' => function() { return $this->config['minSpecial']['enabled']; }
+            ],
+            [
+                'password', ZxcvbnPasswordValidator::className(), 'minScore' => $this->config['zxcvbn']['minScore'],
+                'skipOnError' => true,
```
forevermatt: What is `skipOnError` for?

application/common/models/Password.php
```diff
@@ -0,0 +1,152 @@
...
+
+        return $password;
+    }
+
+    /**
+     * @param int $userId
+     * @throws BadRequestHttpException
+     * @throws ServerErrorHttpException
+     */
+    public function save($userId)
```
forevermatt: Would it make more sense for the `$userId` to be set when the Password object is instantiated rather than when `save()` is called? It seems like a Password should immutably be linked to a specific user / user id, rather than not knowing the user id until save is called. 

What if we changed the Password object to require the `$userId` in the constructor (and save it to a private property), and not provide any way to change it without creating a new Password object?

application/common/models/Reset.php
```diff
@@ -487,23 +484,23 @@ public function setType($type, $methodId = null)
         if (in_array($type, [self::TYPE_SPOUSE, self::TYPE_SUPERVISOR, self::TYPE_PRIMARY])) {
             $this->type = $type;
             $this->method_id = null;
-        } elseif ($type == self::TYPE_METHOD) {
+        } elseif (in_array($type, [self::TYPE_METHOD, Method::TYPE_EMAIL, Method::TYPE_PHONE])) {
```
forevermatt: Are we intentionally comparing the new Reset `$type` both to a constant in the Reset class and to constants in the Method class? Something seems odd about that...

application/common/models/User.php
```diff
@@ -350,4 +376,56 @@ public function getVerifiedMethods()
     {
         return Method::findAll(['user_id' => $this->id, 'verified' => 1]);
     }
+
+    public function getPasswordMeta()
+    {
+        $thisUser = self::findOne(['id' => $this->id]);
```
forevermatt: Is there a reason we can't just use `$this` here instead of having to retrieve a User object that has the id that this User object has?

application/common/models/Password.php
forevermatt: It may also be worth using `$employeeId` rather than `$userId` within the Password class, since that is what the PasswordStoreInterface uses, as well as to avoid confusion of which property to use when calling this from the User class.

application/common/models/User.php
```diff
@@ -350,4 +376,56 @@ public function getVerifiedMethods()
...
+    public function createAccessToken($clientId)
+    {
+        /*
+             * Create access_token and update user
+             */
+        $accessToken = Utils::generateRandomString(32);
+        /*
+         * Store combination of clientId and accessToken for bearer auth
+         */
+        $this->access_token = $clientId . $accessToken;
```
forevermatt: These are pretty minor issues, but it may be worth cleaning this function up a little. For example: 
- Does an access token include the client id? Lines 417/419 and 421 seem to give different answers to that question.
- The first comment should probably be moved out to be the doc. comment's description of the function as a whole. That would also resolve the problem of there being no indication from the "outside" (aka. the function's name and doc. comment) that the function is both creating _and saving_ the access token. That may not matter... but more clarity is almost always helpful.

application/frontend/controllers/AuthController.php
```diff
@@ -128,6 +164,65 @@ public function getAfterLoginUrl($returnTo)
...
+     * @return string
+     * @throws BadRequestHttpException
+     */
+    public function getClientIdOrFail()
+    {
+        $clientId = \Yii::$app->request->get('client_id');
+        if ($clientId === null) {
+            $clientId = \Yii::$app->session->get('clientId');
+            if ($clientId === null) {
+                throw new BadRequestHttpException('Missing client_id');
```
forevermatt: Does the `'Missing client_id'` text get returned to the requesting browser? If so, do we want to tell users making invalid requests why their request was invalid? It seems like invalid requests will (almost always) be coming from users trying to hack the system, rather than users using it the normal way, in which case it seems better to not give them any helpful clues.

application/frontend/controllers/AuthController.php
```diff
@@ -128,6 +164,65 @@ public function getAfterLoginUrl($returnTo)
...
+            }
+        }
+        \Yii::$app->session->set('clientId', $clientId);
```
forevermatt: Something about the order in which we handle the client id here (check the request, then check the session, then if we found it in either place we store it back into the session) seems off to me, but I can't quite put my finger on why. Anyone else have any misgivings about this part of the code?

application/frontend/controllers/AuthController.php
```diff
@@ -128,6 +164,65 @@ public function getAfterLoginUrl($returnTo)
...
+
+        return $clientId;
+    }
+
+    /**
+     * Get state from request or session and then store in session
```
forevermatt: What does 'state' refer to here?

application/frontend/controllers/AuthController.php
```diff
@@ -86,35 +117,40 @@ public function actionLogin()
 
     public function actionLogout()
     {
-        if (\Yii::$app->user->isGuest) {
+        $accessToken = \Yii::$app->request->get('access_token');
+        if ($accessToken !== null) {
```
forevermatt: It looks like the user will only be logged out if their logout HTTP request included an access_token. Is that intentional?

application/common/models/Password.php
fillup: This way if a different validation filter errors it will not call zxcvbn api.

application/common/models/Password.php
fillup: Yeah those sound like good ideas. I refactored it now to pass employeeId and newPassword into the Password::create() method.

application/common/models/Reset.php
fillup: Yeah it might seem a bit weird. Basically we have these "meta" types of supervisor, spouse, and primary, but then they can also be the preconfigured options and when one of the preconfigured options we also need their unique id.

application/frontend/controllers/AuthController.php
fillup: We don't use sessions for identity/authorization anymore, just the token, so the logout method needs the token to nullify the token in the db and redirect to IdP to kill session there too.

forevermatt: Hmm, okay. Do we use sessions for anything?

forevermatt: I'm just thinking that if we aren't using sessions for anything, do we want to try to turn off sessions somehow in the PHP settings. If we are using sessions for something (even just some sort of cache), it may still be worth killing the session when they log out.

PR #39, opened 2016-07-21
Adding cron support

fillup:
Currently adding a cron job to retry emails that get queued up. It will attempt to retry the oldest five emails every minute. Emails are only queued if initial attempt to send fails, so far in testing we have not seen one fail so this is a backup process just in case.

### DIFF
--- a/application/console/controllers/CronController.php
+++ b/application/console/controllers/CronController.php
@@ -1,0 +1,43 @@
+<?php
+namespace console\controllers;
+
+use common\models\EmailQueue;
+use yii\console\Controller;
+
+class CronController extends Controller
+{
+    /**
+     * Retry sending emails from queue in small batch sizes
+     */
+    public function actionSendQueuedEmail()
+    {
+        try {
+            echo "starting cron/send-queued-email" . PHP_EOL;
+
+            $batchSize = \Yii::$app->params['emailQueueBatchSize'];
+            $queued = EmailQueue::find()->orderBy(['last_attempt' => SORT_ASC])->limit($batchSize)->all();
+
+            if ($queued === null) {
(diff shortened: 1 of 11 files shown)


Baggerone:
Looks OK to me.

application/console/controllers/CronController.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    /**
+     * Retry sending emails from queue in small batch sizes
+     */
+    public function actionSendQueuedEmail()
+    {
+        try {
+            echo 'starting cron/send-queued-email' . PHP_EOL;
+
+            $batchSize = \Yii::$app->params['emailQueueBatchSize'];
```
forevermatt: What happens if the `emailQueueBatchSize` param isn't set?

application/console/controllers/CronController.php
```diff
@@ -0,0 +1,43 @@
...
+            $queued = EmailQueue::find()->orderBy(['last_attempt' => SORT_ASC])->limit($batchSize)->all();
+
+            if ($queued === null) {
```
forevermatt: I think `$queued` will be an array, so you'll probably want to check for it to be an empty array rather than be null.

forevermatt:
@fillup Okay, those two comments are the only issues I saw (albeit a bit late) with the PR.

application/console/controllers/CronController.php
fillup: It would error, but currently it is hard coded in the main config so it will always be present.

application/console/controllers/CronController.php
fillup: Thanks, bummer that all() has different behavior from findOne()

PR #51, opened 2016-08-02
error reporting cleanup

fillup:


application/common/models/Method.php
```diff
@@ -131,7 +140,7 @@ public static function createAndSendVerification($userId, $type, $value)
                 throw new BadRequestHttpException($e->getMessage(), $e->getCode());
             }
         } else {
-            $method->value = $value;
+            $method->value = strtolower($value);
```
forevermatt: Should this be `mb_strtolower(...)`? I'm not sure what all `$value` can represent here. (If we're using that string library, it probably also has some to-lower sort of function).

application/common/models/Method.php
```diff
@@ -179,20 +188,64 @@ public static function createAndSendVerification($userId, $type, $value)
...
+        ])->all();
+
+        /** @var Method $existMethod */
+        foreach ($existing as $existMethod) {
+            if ($existMethod->type === self::TYPE_PHONE) {
+                if (Utils::stripNonNumbers($existMethod->value) === Utils::stripNonNumbers($value)) {
+                    $existMethod->sendVerification();
+                    return $existMethod;
+                }
+            } else {
```
forevermatt: To help clarify that the only "else" case (currently) is email, it may be worth doing an `elseif ($existMethod->type === self::TYPE_EMAIL) {` here, possibly with a separate "else" case to throw an error in case we ever add another type and forget to update this.

application/common/models/Method.php
```diff
@@ -131,7 +140,7 @@ public static function createAndSendVerification($userId, $type, $value)
                 throw new BadRequestHttpException($e->getMessage(), $e->getCode());
             }
         } else {
```
forevermatt: Similar to my comment below, it may help readability (and possible future error conditions) to make this 
`} else {` 
an 
`} elseif ($type === self::TYPE_EMAIL) {`
and add a separate else to handle unknown types.

application/common/models/Method.php
fillup: Its an email address, do we need mb_functions for email addresses?

application/common/models/Method.php
```diff
@@ -179,20 +188,64 @@ public static function createAndSendVerification($userId, $type, $value)
...
             $this->verification_code = $this->sendVerificationPhone();
-            if ( ! $this->save()) {
-                throw new \Exception('Unable to save method after sending phone verification', 1461441850);
-            }
         } else {
-            throw new \Exception('Invalid method type', 1461432437);
+            throw new BadRequestHttpException('Invalid method type', 1461432437);
+        }
+
+        if ( ! $this->save()) {
```
forevermatt: As we did somewhere else, should we make sure we successfully save the updated attempts-counter before sending the verification message (to avoid situations where a failure to save the updated attempts-counter could let someone send an unlimited number of verification messages)?

application/common/models/Method.php
fillup: good call

application/common/models/Method.php
fillup: good call

application/common/models/User.php
```diff
@@ -111,9 +126,11 @@ public static function findOrCreate($username = null, $email = null, $employeeId
             $user->idp_username = $personnelUser->username;
             $user->email = $personnelUser->email;
             if ( ! $user->save()) {
-                /*
-                 * add logging with model validation errors
-                 */
+                \Yii::error([
+                    'action' => 'create new user',
+                    'status' => 'error',
+                    'error' => $user->getFirstErrors(),
```
forevermatt: Will providing an array as the value for `'error'` here cause a problem?

application/common/models/Method.php
```diff
@@ -179,20 +188,64 @@ public static function createAndSendVerification($userId, $type, $value)
...
+                    $existMethod->sendVerification();
+                    return $existMethod;
+                }
+            } else {
+                if (strtolower($existMethod->value) === strtolower($value)) {
+                    $existMethod->sendVerification();
+                    return $existMethod;
+                }
+            }
+        }
```
Baggerone: Another possibility?
`$checkFunction = strtolower;
if ($type === self:TYPE_PHONE) { $checkFunction = Utils::stripNonNumbers}
foreach ($existing as $existMethod) {
    if ($checkFunction($existingMethod->value) === $checkFunction($value)) {
        $existMethod->sendVerification();
        return $existMethod;
    }`

application/common/models/User.php
fillup: no, since it gets serialized to json it works, and the email target serializes it with that vardumper class that writes it out nicely too

application/common/models/User.php
```diff
@@ -312,7 +331,7 @@ public static function findIdentity($id)
      * Finds an identity by the given token.
      *
      * @param string $token the token to be looked for
-     * @return User|null the identity object that matches the given token.
+     * @return User|array the identity object that matches the given token.
```
forevermatt: I think this was correct before (as `null`). See http://www.yiiframework.com/doc-2.0/yii-db-activequery.html#one()-detail

application/common/models/Method.php
fillup: yeah I can move things around a bit

application/common/models/User.php
fillup: oh yeah, its ```all()``` that returns an empty array

application/common/models/Method.php
fillup: yeah thats cleaner

PR #53, opened 2016-08-04
Improve some logging and exception handling

fillup:


application/common/models/Method.php
```diff
@@ -210,9 +210,13 @@ public static function checkForExistingAndResend($userId, $type, $value)
         ])->all();
 
 
-        $checkFunction = 'mb_strtolower';
+        $checkFunction = function($value) {
```
forevermatt: As elsewhere, do we want to make it clear that this is for email addresses (both for readability and so that, if we ever are trying this with something other than a phone number / email address, we'll know to look at this code again)?

application/common/models/Method.php
```diff
@@ -364,11 +368,13 @@ public static function deleteExpiredUnverifiedMethods()
                                 ->all();
 
         foreach ($methods as $method) {
-            if ( ! $method->delete()) {
+            try {
+                $method->delete();
```
forevermatt: Does `$method->delete()` ever return a value, or does it exclusively depend on throwing exceptions?

application/common/models/Reset.php
```diff
@@ -648,4 +648,23 @@ public static function deleteExpired()
...
+        if ($this->type === self::TYPE_METHOD) {
+            return $this->method->getMaskedValue();
+        } elseif ($this->type === self::TYPE_PRIMARY) {
+            return Utils::maskEmail($this->user->email);
+        } elseif ($this->type == self::TYPE_SUPERVISOR) {
+            return Utils::maskEmail($this->user->getSupervisorEmail());
+        } elseif ($this->type == self::TYPE_SPOUSE) {
+            return Utils::maskEmail($this->user->getSpouseEmail());
+        } else {
+            return 'Invalid reset type';
```
Baggerone: Is this adequate? If it ever gets hit, it won't give very helpful info.

application/frontend/controllers/ResetController.php
```diff
@@ -210,12 +210,24 @@ public function actionValidate($uid)
         $isValid = $reset->isUserProvidedCodeCorrect($code);
         if ($isValid === true) {
 
+            $ipAddress = Utils::getClientIp(\Yii::$app->request);
+            if ($reset->type === Reset::TYPE_METHOD) {
+                $methodType = $reset->method->type;
+            } else {
```
Baggerone: Unnecessary 'else' ... just kidding :)

application/common/models/Method.php
fillup: It'll return the number of rows affected or an exception if there was an error. It is possible to return zero, but if it cant delete for some reason an exception should be thrown.

application/common/models/Reset.php
fillup: no reason it should ever be hit

application/common/models/Method.php
```diff
@@ -364,11 +386,16 @@ public static function deleteExpiredUnverifiedMethods()
                                 ->all();
 
         foreach ($methods as $method) {
-            if ( ! $method->delete()) {
+            try {
+                $deleted = $method->delete();
+                if ($deleted === 0 || $deleted === false) {
```
fillup: @forevermatt I added a check here for that condition

PR #67, opened 2017-04-26
added env var to enable/disable requiring recaptcha, default is required

fillup:
When running locally for dev without a UI its a pain to deal with recaptcha, this will allow disabling the requirement for it. I added it as a separate var rather than conditionally based on the recaptcha key and secret being present to ensure it is only disabled intentionally, not due to forgetting one of the other vars.


lvail:
Looks like this will be useful in dev but could also be useful in prod if there is trouble with recaptcha service.

PR #69, opened 2017-06-12
Feature/memecache

Baggerone:
Changed configs to use memcache instead of sessions


application/frontend/config/main.php
```diff
@@ -4,6 +4,13 @@
 
 /* Get frontend-specific config settings from ENV vars or set defaults. */
 $frontCookieSecure = Env::get('FRONT_COOKIE_SECURE', true);
+// default to match docker-compose.yml settings
```
fillup: What do you think about using the getArrayFromPrefix approach for arrays of configs like this instead of explicitly getting each parameter as separate vars? I don't know if this is cleaner, but it could allow providing more configuration to each server if needed/supported and would work for one or two memcache servers in configuration. 

```php
$memcache1 = Env::getArrayFromPrefix('MEMCACHE_CONFIG1_');
$memcache2 = Env::getArrayFromPrefix('MEMCACHE_CONFIG2_');
$memcacheConfig = [];
if ( is_array($memcache1) && ! empty($memcache1) {
  $memcacheConfig[] = $memcache1;
}
if ( is_array($memcache2) && ! empty($memcache2) {
  $memcacheConfig[] = $memcache2;
}
```

Then in sessionCache/servers config you can just have:

```php
'servers' => $memcacheConfig
```
I think that'll make it easier to support one or two memcache servers.

application/frontend/config/main.php
```diff
@@ -21,14 +28,32 @@
...
+                    'host' => $memcacheHost2,
+                    'port' => $memcachePort2,
+                    'weight' => $memcacheWeight2,
+                ],
             ],
         ],
+        'session' => [
+            'class' => 'yii\web\CacheSession',
+            'cache' => 'sessionCache',
+//            'cookieParams' => [// http://us2.php.net/manual/en/function.session-set-cookie-params.php
```
fillup: This is still needed right? We need to write a session cookie to the browser for us to be able to find their session in memcache.

PR #72, opened 2017-07-12
Feature/refresh personnel data for user with specific email

fillup:
Quite regularly we have users who have used a positional email account as their personal email address, then assignments change and someone else gets that address. This update:

When checking if a person's information has changed to update their local User record it will check if their new email address is already in use by another user in our database. If it is already in use it will call the personnel system for that other user to refresh their information before updating the current user's record.


application/common/models/User.php
```diff
@@ -199,6 +212,73 @@ public function updateProfileIfNeeded($firstName, $lastName, $username, $email)
     }
 
     /**
+     * In case where email address has been reassigned to $this user, refresh profile for user
+     * with email address currently in case their email address has changed too
+     * @param string $email
+     * @return bool
+     * @throws \Exception
+     */
+    public static function refreshPersonnelDataForUserWithSpecificEmail($email)
```
wcjr: Should we consider other places in the db, e.g., methods, where we should wipe the current associated data?

application/tests/unit/common/models/UserTest.php
```diff
@@ -305,5 +305,35 @@ public function testPasswordChangeLog()
...
+    }
+
+    public function testRefreshPersonnelDataForUserWithSpecificEmail()
+    {
+        /*
+         * Test user who does not have updated information in personnel
+         */
+        $user1 = $this->users('user1');
+        User::refreshPersonnelDataForUserWithSpecificEmail($user1->email);
+        $notupdated = User::findOne(['id' => $user1->id]);
```
forevermatt: I don't quite understand what's going on here and in the lines below, but presumably that's because I don't know the fixture details for these users. Is there a way to make this test more clear? (It may not be worth the effort.)

application/common/models/User.php
```diff
@@ -199,6 +212,73 @@ public function updateProfileIfNeeded($firstName, $lastName, $username, $email)
     }
 
     /**
+     * In case where email address has been reassigned to $this user, refresh profile for user
+     * with email address currently in case their email address has changed too
+     * @param string $email
+     * @return bool
```
fillup: whether or not the profile was updated, I'll add to the comments.

application/common/models/User.php
fillup: yeah, although I'll probably do that as a separate PR to not delay this one

application/common/models/User.php
```diff
@@ -199,6 +212,73 @@ public function updateProfileIfNeeded($firstName, $lastName, $username, $email)
...
+                sprintf('User with email %s not found', $email),
+                1499817281
+            );
+        }
+
+        try {
+            /** @var PersonnelUser $personnelUser */
+            $personnelUser = \Yii::$app->personnel->findByEmployeeId($user->employee_id);
+        } catch (\Exception $e) {
+            if ($e instanceof NotFoundException) {
```
fillup: good catch

application/run-cron.sh
```diff
@@ -17,23 +17,8 @@ chown -R www-data:www-data \
...
-if [[ $rc != 0 ]]; then
-    logger --priority user.err --stderr "Migrations failed with status ${rc} and output: ${output}"
-    exit $rc;
-fi
-
-output=$(/data/yii migrate --interactive=0 --migrationPath=console/migrations-local 2>&1)
-# If they failed, exit.
-rc=$?;
-if [[ $rc != 0 ]]; then
-    logger --priority user.err "Migrations failed with status ${rc} and output: ${output}"
```
fillup: yes, runny is our new wrapper, I didn't mean to include it in this PR but in a different one, but oh well, it can come in at the same time. See https://github.com/silinternational/runny

application/run-cron.sh
```diff
@@ -17,23 +17,8 @@ chown -R www-data:www-data \
...
-    exit $rc;
-fi
-
+runny /data/yii migrate --interactive=0
```
fillup: https://github.com/silinternational/runny

application/tests/unit/common/models/UserTest.php
fillup: unfortunately there is a need to manually maintain common fixture data in the mock/test personnel component and the database fixtures.

application/common/models/User.php
wcjr: hahaha  good `catch` oh you're a funny one!

PR #73, opened 2017-07-13
Updates for IdBroker and reassigning email addresses

fillup:


wcjr:
wow, that's a big one

PR #74, opened 2017-07-19
refactoring to use idpDisplayName whenever displayed to users

fillup:


application/common/helpers/Utils.php
```diff
@@ -206,7 +206,7 @@ public static function getFrontendConfig()
 
         $config = [];
 
-        $config['idpName'] = $params['idpName'];
+        $config['idpName'] = $params['idpDisplayName'];
```
forevermatt: Would it be helpful to change `$config['idpName']` to `$config['idpDisplayName']`, as well (since we're still using `idpName` for the IDP's code name)?

application/tests/unit/common/helpers/UtilsTest.php
```diff
@@ -129,7 +130,7 @@ public function testGetFrontendConfig()
 
         $params = \Yii::$app->params;
         $config = Utils::getFrontendConfig();
-        $this->assertEquals($params['idpName'], $config['idpName']);
+        $this->assertEquals($params['idpDisplayName'], $config['idpName']);
```
forevermatt: As a parallel change to my my comment on Utils::getFrontendConfig(), should we change `$config['idpName']` here to `$config['idpDisplayName']`?

application/common/helpers/Utils.php
fillup: Not in this case, the $config array is returned as the body for the /config API call, so changing the key here changes the API.

application/tests/unit/common/helpers/UtilsTest.php
fillup: same as above

application/common/config/main.php
```diff
@@ -21,7 +21,8 @@
 $fromName = Env::get('FROM_NAME');
 $appEnv = Env::get('APP_ENV');
 $idpName = Env::get('IDP_NAME');
-$idpUsernameHint = Env::get('IDP_USERNAME_HINT', $idpName . ' username, ex: first_last');
+$idpDisplayName = Env::get('IDP_DISPLAY_NAME', $idpName);
+$idpUsernameHint = Env::get('IDP_USERNAME_HINT', $idpDisplayName . ' username, ex: first_last');
```
Baggerone: Do these (and others?) need to be added to the local.env.dist?

fillup: I put it in the `common.env`

PR #77, opened 2017-07-24
Feature/email service for verifications

fillup:
I'll update the composer dependency to be a version number after the client is released.

### DIFF
--- a/application/common/models/Verification.php
+++ b/application/common/models/Verification.php
@@ -65,17 +65,25 @@ class Verification extends Model
          * local EmailQueue service
          */
         if (\Yii::$app->params['emailVerification']['useEmailService']) {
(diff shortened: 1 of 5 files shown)


application/common/models/Verification.php
```diff
@@ -58,16 +60,57 @@ public static function sendEmail(
...
-            $ccAddress,
-            $eventLogUserId,
-            $eventLogTopic,
-            $eventLogDetails
-        );
+        /*
+         * If configured to use external email service send through that instead of using
+         * local EmailQueue service
+         */
+        if (\Yii::$app->params['emailVerification']['useEmailService']) {
```
Baggerone: You could check for the else condition first (which has a shorter code block) and have a return at the end of it and then not have this long if block indented.

PR #84, opened 2017-11-09
Adding 2-Step Verification support

fillup:


application/frontend/controllers/MfaController.php
```diff
@@ -0,0 +1,100 @@
...
+                        'roles' => ['@'],
+                    ],
+                ]
+            ]
+        ]);
+    }
+
+    public function init()
+    {
+        parent::init(); // TODO: Change the autogenerated stub
```
forevermatt: Does this `TODO` still need to be addressed?

application/frontend/controllers/MfaController.php
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        try {
+            if ($this->idBrokerClient->mfaVerify($mfaId, \Yii::$app->user->identity->employee_id, $value)) {
+                \Yii::$app->response->statusCode = 204;
+                return;
+            }
+            throw new BadRequestHttpException('Invalid code provided');
+        } catch (\Exception $e) {
+            throw new BadRequestHttpException($e->getMessage());
```
forevermatt: Rather than a `BadRequestHttpException`, should this be a server error exception of some sort (for an HTTP 500 response)?

forevermatt: Also, are we sure we want to return the message of an unknown exception to the end user's browser? I know we (at least sometimes) sanitize those, logging the full error message but sending something "safe" to the end user.

application/common/config/local.php.dist
```diff
@@ -1,23 +1,76 @@
...
+//            ['class' => 'Sil\IdpPw\Common\Personnel\IdBroker\IdBroker'],
+//            Env::getArrayFromPrefix('ID_BROKER_')
+//        ),
         'auth' => [
-            'class' => '\Sil\IdpPw\Auth\Saml',
-            'signRequest' => true,
-            'checkResponseSigning' => true,
-            'requireEncryptedAssertion' => true,
+            'class' => Sil\IdpPw\Auth\Saml::class,
+            'signRequest' => false,
```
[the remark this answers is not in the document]
fillup: I had it disabled by default in dev for troubleshooting, in staging and prod it is all enabled.

application/frontend/controllers/MfaController.php
fillup: oh, no, that was an ide thing I didnt clean up

application/frontend/controllers/MfaController.php
fillup: yeah that is a good idea, I'll make those changes

fillup: How is that update?

PR #90, opened 2017-11-27
Latest develop changes

Baggerone:


Makefile
```diff
@@ -3,9 +3,11 @@ start: api
 test:
 	make testunit && make testapi
 
-testunit: composer rmTestDb upTestDb yiimigratetestDb yiimigratetestDblocal rmTestDb
+testunit: composer emailcron rmTestDb upTestDb yiimigratetestDb yiimigratetestDblocal
+	docker-compose run emailcron whenavail emaildb 3306 100 ./yii migrate --interactive=0
```
forevermatt: This line seems redundant, since the new `emailcron` entry in the line above will automatically run `run-cron.sh`, which runs the database migrations (albeit without the use of `whenavail`).

application/frontend/controllers/MfaController.php
```diff
@@ -0,0 +1,108 @@
...
+        if ($value === null) {
+            throw new BadRequestHttpException('Value is required');
+        }
+
+        try {
+            if ($this->idBrokerClient->mfaVerify($mfaId, \Yii::$app->user->identity->employee_id, $value)) {
+                \Yii::$app->response->statusCode = 204;
+                return;
+            }
+            throw new BadRequestHttpException('Invalid code provided');
```
forevermatt: I think this may be a minor problem (not blocking, but worth addressing after this release). This `BadRequestHttpException` (aka. `HTTP 400`, I think) is immediately caught and instead a `ServerErrorHttpException` (aka. `HTTP 500` is thrown. We may want to move this `throw new BadRequestHttpException('Invalid code provided');` line down below the `catch` block.

fillup: What about in the catch block starting with:
```php
if ($e instanceof BadRequestHttpException) {
     throw $e;
}
```

fillup: then we wont log and wont throw the 500 if it was just an invalid code.

PR #109, opened 2018-09-20
Integrate Personnel\IdBroker into this repo

Schparky:
- Pulled in the code from idp-pw-api-personnel-id-broker and removed that dependency.
- Fixed local testing to not overwrite information in common/config/local.php
- Moved IdBroker unit tests from idp-pw-api-personnel-id-broker into this repo
- Added a broker instance to interact with the unit and api tests
- Also brought in Auth code from idp-pw-api-auth-saml


README.md
```diff
@@ -66,35 +66,34 @@ With the goal of being reusable, this application is developed with a component
...
 * Example implementation: [idp-pw-api-passwordstore-ldap](https://github.com/silinternational/idp-pw-api-passwordstore-ldap)
 
 ### Personnel Component
-The personnel component is used to look up informaton about users from your companies personnel system. This includes verifying that they are an active employee, getting information about them like name, email, employee id, if they have a supervisor and what their supervisors email address is and if the personnel system is aware of spouses it can also provide the spouse's email address.
+The personnel component is used to look up informaton about users from your company's personnel system. This includes verifying that they are an active employee, getting information about them like name, email, employee id, whether they have a supervisor and what their supervisors email address is. If the personnel system is aware of spouses it can also provide the spouse's email address.
 
 * Component ID: ```personnel```
-* Implement interface: ```Sil\IdpPw\Common\Personnel\PersonnelInterface```
+* Implement interface: ```common\components\Personnel\PersonnelInterface```
```
forevermatt: Since namespaces are global, we should probably have the new namespace begin with something like `\Sil\IdpPw\common\components\Personnel\...` (rather than merely using `\common\components\Personnel\..`).

application/frontend/web/index.php
```diff
@@ -12,7 +12,7 @@
      * Load environment config if present, else expect local.php
      */
     $appEnv = \Sil\PhpEnv\Env::get('APP_ENV', 'production');
-    if (! in_array($appEnv, ['development', 'dev', 'staging', 'stage', 'stg', 'production', 'prod', 'prd'])) {
+    if (! in_array($appEnv, ['test', 'development', 'dev', 'staging', 'stage', 'stg', 'production', 'prod', 'prd'])) {
```
forevermatt: Not directly related to this change, but it seems to generally be a good practice when using [in_array(...)](http://php.net/in_array) to use this form...
```php
in_array($needle, $haystack, true)
```
... supplying that last `true` to force a strict comparison, rather than this form...
```php
in_array($needle, $haystack)
```
... because the latter can "match" on a boolean true (because `true == 'any non empty string'`, but `true !== 'any string'`).

Schparky: Good suggestion, thanks.

README.md
Schparky: As discussed, since this is not a repo that could be easily integrated with another project, I'll leave this as is.

application/common/config/test.php
```diff
@@ -52,9 +52,9 @@
                 'host' => null,
             ],
         ],
-        'personnel' => [
-            'class' => 'tests\mock\personnel\Component',
-        ],
+//        'personnel' => [
```
forevermatt: Particular reason these lines were commented out?

Schparky: Yes, to revert the default of using the Broker container rather than the mocks. I chose to leave those lines so that it's easy to use the mocks if needed. In reality, I could have completely removed the Personnel mocks, and might still do that.

PR #113, opened 2018-09-28
Integrate Password Store tests

Schparky:
Brought in PHPUnit tests from:
- idp-pw-api-passwordstore-id-broker
- idp-pw-api-passwordstore-ldap

and Behat tests from:
- idp-pw-api-passwordstore-multiple

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 start: api
 
 test:
-	make testunit && make testapi
+	make testunit && make testapi && make testbehat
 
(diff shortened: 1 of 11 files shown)


Schparky:
@forevermatt There is a Scrutinizer issue pointing to line 90 of common/components/passwordStore/IdBroker.php. It is saying that setPassword always returns null, but is assigned to $update. By looking at the code, I don't understand how it works, but it does. Could you walk through it with me on Monday?

PR #114, opened 2018-10-01
Integrate phone verification from idp-pw-api-phoneverification-nexmo

Schparky:
Integrated code from idp-pw-api-phoneverification-nexmo, and corresponding interface and exception from idp-pw-api-common. This is the last of the external packages in this batch of integrations.

### DIFF
--- a/application/common/components/phoneVerification/Sms.php
+++ b/application/common/components/phoneVerification/Sms.php
@@ -52,7 +52,7 @@ class Sms extends Base implements PhoneVerificationInterface
         } catch (RequestException $e) {
             if ($e->hasResponse()) {
                 $response = $e->getResponse();
-                $body = json_decode($response->getBody());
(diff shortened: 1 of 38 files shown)


application/common/components/phoneVerification/Sms.php
```diff
@@ -52,7 +52,7 @@ public function send($phoneNumber, $code)
         } catch (RequestException $e) {
             if ($e->hasResponse()) {
                 $response = $e->getResponse();
-                $body = $response->json();
+                $body = json_decode($response->getBody());
```
forevermatt: You may need to add a 2nd argument to json_decode (set to `true`) so that the results will be an associative array rather than an object.

Typically when using json_decode, we do so like this:
```php
$body = json_decode($response->getBody(), true);
```

PR #119, opened 2018-10-10
Limit access if auth level is not login

Schparky:
MethodController and MfaController now disallow all actions if the auth scope (auth_type) is not a full login, returning a http 403 response.

Updated api tests accordingly, and added new tests to confirm 403 response.

### DIFF
--- a/application/frontend/controllers/MfaController.php
+++ b/application/frontend/controllers/MfaController.php
@@ -25,7 +25,10 @@ class MfaController extends BaseRestController
                 'rules' => [
                     [
                         'allow' => true,
-                        'roles' => ['@'],
+                        'matchCallback' => function () {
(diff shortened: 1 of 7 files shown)


application/frontend/controllers/MfaController.php
```diff
@@ -25,7 +25,10 @@ public function behaviors()
                 'rules' => [
                     [
                         'allow' => true,
-                        'roles' => ['@'],
+                        'matchCallback' => function () {
+                            $user = \Yii::$app->user->identity;
+                            return ( $user->isAuthScopeFull());
```
forevermatt: This is a very minor issue, but you don't actually need the parentheses around this function call before returning the value. You can simply use...
```php
return $user->isAuthScopeFull();
```

PR #122, opened 2018-10-17
error message i18n

Schparky:
Add internationalization support for error messages:
* Uses Yii2 `ContentNegotiator` filter to automatically set the language based on `Accept-language` header received.
* Wrapped all text send to `BadRequestHttpException` (400) and `NotFoundHttpException` (404) with `\Yii::t()` translate method.
* Applied numerous Scrutinizer auto-fixes. (my apologies for cluttering this PR...should have merged these separately)


application/frontend/config/main.php
```diff
@@ -13,7 +13,17 @@
...
     'basePath' => dirname(__DIR__),
-    'bootstrap' => ['log', 'errorHandler'],
+    'bootstrap' => [
+        'log',
+        'errorHandler',
+        [
+            'class' => 'yii\filters\ContentNegotiator',
+            'languages' => [
+                'en',
+                'fr'
```
forevermatt: I know we sometimes include Spanish and Korean as well. Did we intentionally leave this at only English and French?

application/common/components/phoneVerification/Base.php
```diff
@@ -112,8 +112,12 @@ public function format($phoneNumber)
                 return sprintf('%s %s', $insights['country_prefix'], $insights['national_format_number']);
             } else {
                 throw new \Exception(
-                    'Unable to verify phone number for formatting, please check the number and try again. ' .
-                    'Error code: ' . $insights['status'],
+                    \Yii::t(
+                        'app',
+                        'Unable to verify phone number for formatting, please check the number and try again. ' .
+                            'Error code: {code}',
```
forevermatt: This is a minor, somewhat subjective comment, but we typically don't do a full indent on a string-continuation. So rather than this...
```php
/Yii::t(
    'app',
    'A rather long string that ' .
        'continues on a 2nd line.'
)
```
... we might do this:
```php
/Yii::t(
    'app',
    'A rather long string that '
    . 'continues on a 2nd line.'
)
```
That seems to make it easier to tell (without having to go hunting for the end of the line) that it's a continuation of the string on the previous line and not some further function call causing an additional indentation level.

application/frontend/config/main.php
Schparky: No good reason. It's just a placeholder for now, as I don't know which languages will be supported. Also need some discussion about how best to generate the translation files.

application/common/components/phoneVerification/Base.php
Schparky: Thanks. I'll make that adjustment.

PR #125, opened 2018-10-24
Integrate Google passwordStore component

Schparky:
Pulled in code, tests, and readme from idp-pw-api-passwordstore-google.


Schparky:
@forevermatt, there is a Scrutinizer [warning](https://scrutinizer-ci.com/g/silinternational/idp-pw-api/inspections/e0337621-678e-49b3-96f3-5152aaa7d6bb/issues/files/application/common/components/passwordStore/Google.php?status=failed&conditionIds%5B0%5D=312699) that doesn't make sense to me. Can you tell me if it's safe to ignore?

Schparky:
@forevermatt, I tried to remove that dynamic class reference, but found that it was used in one of the Behat tests to insert dummy user data. Rather than try to implement that some other way, I reverted the change and hid the Scrutinizer warning.

PR #126, opened 2018-10-26
Remove phone verification

Schparky:
Removed phone (SMS) verification capability, retaining just enough logic to handle existing records in the `method` table.


Schparky:
As we discussed a moment ago, when the entire `method` table and associated code is moved to ID Broker, a script will be created to move the data to ID Broker. After that script has completed successfully, the `method` table will be dropped from the api database.

PR #130, opened 2018-11-08
Transition to use of broker-based recovery methods table

Schparky:
This is the next step in transitioning to storing recovery methods in the broker database. This makes it functional with the following caveats:
* may leave some corner cases or error conditions not handled correctly
* the obsolete method table and related code remain, but will be removed later
* test coverage is reduced; new test doubles are needed

Note: when a `Reset` is changed to `TYPE_METHOD`, the email address is cached in the reset table, reducing calls to Broker.


api.raml
```diff
@@ -0,0 +1,360 @@
...
+      last_changed: string
+      expires: string
+    example:
+      {
+        "last_changed": "2018-09-13T19:21:57+00:00",
+        "expires": "2020-09-13T23:59:59+00:00"
+      }
+  Reset:
+    description: |
+      An object holding the emails that can be used for performing
```
fillup:
```suggestion
      An object holding the verified emails that can be used for performing
```

application/common/models/Method.php
```diff
@@ -324,4 +346,39 @@ public static function deleteExpiredUnverifiedMethods()
...
+     * @param string $employeeId
+     * @return null|String[]
+     * @throws \yii\web\NotFoundHttpException
+     * @throws \Exception
+     */
+    public static function getOneVerifiedMethod($uid, $employeeId)
+    {
+        $method = new Method;
+        try {
+            return $method->idBrokerClient->getMethod($uid, $employeeId);
```
fillup: If broker returns unverified methods now this function either needs to be renamed or updated to check the verified status before returning it.

application/common/models/Method.php
```diff
@@ -324,4 +346,39 @@ public static function deleteExpiredUnverifiedMethods()
...
 
+    /**
+     * Gets all verified methods for user specified by $employeeId
+     * @param string $employeeId
+     * @return String[]
+     */
+    public static function getVerifiedMethods($employeeId)
+    {
+        $method = new Method;
+        return $method->idBrokerClient->listMethod($employeeId);
```
fillup: If broker returns unverified methods now this'll need updating too

application/common/models/Method.php
Schparky: @wcjr suggested leaving this alone since it isn't used in the UI, deferring the decision on whether to totally remove it.

application/common/models/Method.php
Schparky: Updated.

PR #132, opened 2018-11-20
Improve exception handling

Schparky:
- Add exception handling for all known 400-level status codes on `method` endpoints.
- Fill in all known responses in RAML file
- Clean up 'verified' references `getMethodsAndPersonnelEmails()`

### DIFF
--- a/application/common/models/Method.php
+++ b/application/common/models/Method.php
@@ -350,7 +350,7 @@ class Method extends MethodBase
      * Gets all methods for user specified by $employeeId
      * @param string $employeeId
      * @return String[]
-     * @throws BadRequestHttpException
+     * @throws ServerErrorHttpException
      * @throws ServiceException
      */
     public static function getMethods($employeeId)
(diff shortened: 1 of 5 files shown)


application/common/models/Method.php
```diff
@@ -350,19 +350,30 @@ public static function deleteExpiredUnverifiedMethods()
...
     public static function getMethods($employeeId)
     {
         $method = new Method;
-        return $method->idBrokerClient->listMethod($employeeId);
+
+        try {
+            return $method->idBrokerClient->listMethod($employeeId);
+        } catch (ServiceException $e) {
+            if ($e->httpStatusCode === 400) {
+                throw new BadRequestHttpException(\Yii::t('app', 'Error locating personnel record'), 1542752270);
```
wcjr: would this actually be a 404?

wcjr: the wording through me off a little, sounded like a message when the user wasn’t found but the idea your describing does sound more like a 500 I think

PR #133, opened 2018-11-28
Add "Do Not Disclose" mode

Schparky:
Implemented as a new column on `user` table, and a new `PUT /user` API endpoint.
If `do_not_disclose` is `true`:
- `POST /reset` will respond with 404 status code even if `username` is valid
- password recovery emails will be sent to the user's primary email address and any verified password recovery email addresses


Makefile
```diff
@@ -22,7 +22,7 @@ dockerpullall:
 	docker pull phpmyadmin/phpmyadmin:latest
 	docker pull silintl/data-volume:latest
 	docker pull silintl/mariadb:latest
-	docker pull silintl/php7:latest
+	docker pull silintl/php7:7.2
```
forevermatt: It turns out that there is also a `docker-compose pull` that will automatically pull the relevant images, so this whole make command could be simplified to just that, if desired.

api.raml
```diff
@@ -105,7 +105,31 @@ types:
...
+      contain all of the properties in the personnel record, and may contain other
+      non-personnel information.
+    type: object
+    properties:
+      first_name: string
+      last_name: string
+      idp_username: string
+      email: string
+      password_meta: PasswordMeta
+      do_not_disclose: boolean
```
forevermatt: Will `do_not_disclose` be a mandatory field in the response?

application/common/models/MethodBase.php
```diff
@@ -45,14 +45,14 @@ public function rules()
             [['value'], 'string', 'max' => 255],
             [['verification_code'], 'string', 'max' => 64],
             [['uid'], 'unique'],
-            [['user_id', 'type', 'value'], 'unique', 'targetAttribute' => ['user_id', 'type', 'value'], 'message' => 'The combination of User ID, Type and Value has already been taken.'],
+            [['user_id', 'type', 'value'], 'unique', 'targetAttribute' => ['user_id', 'type', 'value']],
```
forevermatt: Were this file's changes auto-generated by Yii?

Makefile
Schparky: With that knowledge, I think I'll remove that target.

application/common/models/Reset.php
```diff
@@ -164,6 +164,11 @@ public function send()
          */
         $this->trackAttempt('send');
 
+        if ($this->user->do_not_disclose) {
+            $this->sendAll();
```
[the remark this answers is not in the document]
forevermatt: Re: Exceptions - I was thinking of catching any exceptions, probably logging them, then just returning. Something roughly like this:
```php
        if ($this->user->do_not_disclose) {
            try {
                $this->sendAll();
            } catch (\Throwable $t) {
                \Yii::log('Exception during password reset for undisclosed account: ' . $t->getMessage());
            }
            return;
        }
```
The thought being simply to make it more likely that, whatever happens, submitting a password reset for a do_not_disclose account looks like submitting a password reset for a missing account.

application/frontend/controllers/ResetController.php
```diff
@@ -109,7 +109,10 @@ public function actionCreate()
                 'status' => 'error',
                 'error' => 'user not found',
             ]);
-            return new \stdClass();
+            throw new NotFoundHttpException(
+            \Yii::t('app', 'User not found'),
```
forevermatt: Looks like we lost a bit of indentation here.
```suggestion
                \Yii::t('app', 'User not found'),
```

application/tests/unit/common/models/ResetTest.php
```diff
@@ -234,7 +234,29 @@ public function testSendMethodEmail()
         $this->assertTrue(EmailUtils::hasEmailFileBeenCreated($this->tester, 'email-1456769679@domain.org'));
         $this->assertTrue(EmailUtils::hasEmailFileBeenCreated($this->tester, 'requested a password change for their'));
         $this->assertEquals($attempts + 1, $reset->attempts);
+    }
+
+    public function testSendDoNotDisclose()
+    {
+        $this->markTestSkipped('test is broken because fake methods are not accessible in this context');
```
forevermatt: I'm not sure I understand... what was blocking this test from working?

application/common/models/Reset.php
```diff
@@ -272,7 +272,7 @@ private function sendAll()
 
         if (is_iterable($methods)) {
             foreach ($methods as $method) {
-                if ($method['verified']) {
+                if ($method['verified'] ?? false) {
```
forevermatt: At this point, would it be worth extracting a `Method::getVerifiedMethods(...)` function?

application/tests/unit/common/models/ResetTest.php
Schparky: There is a `_before` method that inserts a bunch of fake data into the broker test container. Problem is, there's no easy way to verify the inserted methods. Without that, this test won't send an email to the fake method. I either need to write some code that will receive the verification code and submit that for verification, or I need to use a mocked broker instead of a container. I intend to do the latter.

application/common/models/Reset.php
Schparky: Sure. Just for my own edification, does the idea come from seeing the nested `if`/`foreach`/`if`?

forevermatt: To some extent, yes. It just seemed like there was enough logic in this function that it seemed like some of it could be extracted and significantly simplify this method, to something like this:
```php
    /**
     * Send reset messages to primary email and all verified recovery methods.
     */
    private function sendAll()
    {
        $this->sendPrimary();

        $methods = Method::getVerifiedMethods($this->user->employee_id);

        foreach ($methods as $method) {
            $this->sendMethod($method['value']);
        }
    }
```
Maybe it was that it seemed like this method had more than 1 responsibility: it both had to send resets to the appropriate methods, and it had to determine whether a method was verified or not.

application/common/models/Method.php
```diff
@@ -368,6 +368,30 @@ public static function getMethods($employeeId)
         }
     }
 
+    /**
+     * Gets all verified methods for user specified by $employeeId
+     * @param string $employeeId
+     * @return string[]
```
forevermatt: Is `string[]` the right return type? I would have thought it would either return instances of some class, or possibly arrays.

PR #136, opened 2018-12-04
Add MFA label update

Schparky:
Created a new PUT /mfa/{mfaId} endpoint to change the label of existing MFA options.

### DIFF
--- a/application/frontend/controllers/MfaController.php
+++ b/application/frontend/controllers/MfaController.php
@@ -3,11 +3,15 @@
 namespace frontend\controllers;
 
 use frontend\components\BaseRestController;
+use Sil\Idp\IdBroker\Client\exceptions\MfaRateLimitException;
 use Sil\Idp\IdBroker\Client\IdBrokerClient;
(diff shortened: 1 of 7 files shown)


application/frontend/controllers/MfaController.php
```diff
@@ -95,18 +129,53 @@ public function actionVerify($mfaId)
...
+    /**
+     * @param $mfaId
+     * @throws NotFoundHttpException
+     * @throws ServiceException
+     */
+    public function actionUpdate($mfaId)
+    {
+        $label = \Yii::$app->request->getBodyParam('label');
+        if ($label === null) {
+            return;
```
wcjr: should this throw some kind of `400`?

Schparky: I didn't think so. With a high-level view of `PUT`, it only changes properties that are present. Do you think it should be required?

wcjr: no, I don't think it needs to be required.  I'm just questioning the `return`.  What does the response look like when `label = null`?

PR #137, opened 2018-12-06
Change 'do_not_disclose' to 'hide'

Schparky:
- Rename `do_not_disclose` to `hide`
- Change type from `bool`/`tinyint(1)` to `string` with enum values 'yes' and 'no'


application/common/models/Reset.php
```diff
@@ -164,14 +164,14 @@ public function send()
...
-        if ($this->user->do_not_disclose) {
+        if ($this->user->hide === 'yes') {
             try {
                 $this->sendAll();
             } catch (\Throwable $t) {
                 $log = [
                     'action' => 'reset send all',
                     'status' => 'error',
-                    'error' => 'Exception during password reset for undisclosed account. ' . $t->getMessage()
+                    'error' => 'Exception during password reset for account with hide flag. ' . $t->getMessage()
```
forevermatt: Very minor suggestion, but this could also be phrased "for hidden account":
```suggestion
                    'error' => 'Exception during password reset for hidden account. ' . $t->getMessage()
```

Schparky: I think the benefit of consistent search text was on my mind. As such 'hidden' wouldn't be found when searching for 'hide'.

application/tests/unit/common/models/UserTest.php
```diff
@@ -157,7 +157,9 @@ public function testUpdateProfileIfNeeded()
         $changed = $user->updateProfileIfNeeded($personnelData);
         $this->assertTrue($changed);
 
-        $personnelData->doNotDisclose = ! $personnelData->doNotDisclose;
+        $personnelData->hide === 'no'
```
forevermatt: I'm not used to seeing assignments done in the right-hand sections of a ternary operator, so I'm wondering about readability here. What would you think of something like this instead?
```php
$personnelData->hide = ($personnelData->hide === 'no') ? 'yes' : 'no';
```
I'm tempted to recommend a new `getOppositeOf()` method, but that may be overkill. 🙂

application/common/models/Reset.php
forevermatt: Ahh, that makes sense. Feel free to leave it as you have it.

application/tests/unit/common/models/UserTest.php
Schparky: Yeah, that might be a little overkill. I did take your suggested rewrite, though.

PR #140, opened 2018-12-18
Add 410 response code for expired verification

Schparky:
A new feature of idp-id-broker is to restart verification of expired, unverified password recovery methods. The response code for this event is 410. Now passing this code through to the password API.

Also changed two API endpoints for consistency.
- `PUT /method/{id}` becomes `PUT /method/{id}/verify`
- `POST` becomes `PUT` on `/mfa/{id}/verify`


application/frontend/controllers/MethodController.php
```diff
@@ -125,12 +126,25 @@ public function actionCreate()
...
+    public function actionVerify($uid)
     {
+        $messages = [
+            400 => 'Invalid verification code',
+            404 => 'Recovery method not found',
+            410 => 'Expired verification code',
+            429 => 'Too many failures for this recovery method',
+        ];
+
+        $codes = [
```
wcjr: slick approach!

PR #141, opened 2019-01-02
Renew reset and invite code

Schparky:
Auto-renew expired reset and invite codes. In response to `/reset/{id}/validate` an expired code will result in a 410 response. In response to `/auth/login?invite={code}` an expired code will result in a redirect to an error page.
- a new code is created
- expiration is extended
- new email is sent to user

Note: prior to this PR, reset expiration was ignored except for lazy delete when a new reset is created


application/common/models/User.php
```diff
@@ -636,13 +634,20 @@ public function beforeSave($insert): bool
...
+     * @throws \Sil\Idp\IdBroker\Client\ServiceException
      */
     public static function getUserFromInviteCode(string $inviteCode)
     {
-        /** @var IdBrokerClient $client */
-        $client = \Yii::$app->passwordStore->getClient();
+        /** @var IdBroker $passwordStore */
+        $passwordStore = \Yii::$app->passwordStore;
+
+        if (! (method_exists($passwordStore, 'getClient'))) {
```
forevermatt: It may very well be overkill, but do you think it would be worth defining an interface for Password Stores that support the new user invite? That could make this if-condition more readable (`if ($passwordStore implements NewUserInviteInterface) { ... }`) rather than having to know that the presence of a `getClient()` method means that it supports new user invites.

application/frontend/controllers/AuthController.php
```diff
@@ -231,12 +244,28 @@ protected function getReturnTo()
         return $returnTo;
     }
 
+    /**
+     * Get a return-to url for where to send browser in the event of an error
+     * If it's a relative url (starting with '/') it will be prefixed with uiUrl
+     * @return array|mixed|string
```
forevermatt: Wouldn't this always return a string?

application/common/models/Reset.php
```diff
@@ -331,12 +331,13 @@ private function sendEmail($toAddress, $subject, $view, $ccAddress = null)
     /**
      * Check if user provided code is valid
```
forevermatt: Does this function actually check whether the code is valid (which sounds like it would require it to both be correct and be unexpired) or does it just check whether the code is correct (as the function name itself suggests)?

forevermatt: Given the way we're using this method, it looks like it should just check whether the code is correct, since we need to be able to do that separately from checking whether the code is expired (in order to only restart expire password resets if a valid code was provided).

application/tests/api/ResetCest.php
```diff
@@ -137,6 +137,13 @@ public function test8(ApiTester $I)
         $I->seeResponseCodeIs(200);
     }
 
+    public function test81(ApiTester $I)
+    {
+        $I->wantTo('check response when making unauthenticated PUT request to validate an expired reset code');
```
forevermatt: I'm not sure which this is doing, but it may be worth testing our handling of both an expired reset when given the correct code and an expired reset when given an incorrect code.

api.raml
```diff
@@ -189,16 +189,20 @@ types:
...
           description: New user invite code
           type: string
       responses:
         302:
           description: >
-            Redirection to `ReturnTo` with `state`, `token_type`, `expires_utc` and
-            `access_token` appended as query string parameters.
+            On successful login or invite, a redirection to `ReturnTo` with `state`, `token_type`,
+            `expires_utc` and `access_token` appended as query string parameters. On error, a
+            redirection to `ReturnToOnError` with no query string parameters.
```
forevermatt: This makes it sound like the ReturnToOnError applies both to resets and to invites, but based on the code it looks like it only applies to invites. I'm not sure which was intended.

application/common/models/Reset.php
Schparky: Maybe I'm not understanding correctly, but the method does as the name suggests...checks whether the code matches. It also tracks attempts, but that's unrelated to expiration.

application/frontend/controllers/AuthController.php
Schparky: Yep. Looks like this was auto-generated.

application/tests/api/ResetCest.php
Schparky: I added another test for this.

api.raml
Schparky: I clarified the description.

PR #142, opened 2019-01-03
Add user UUID to `user/me` response

Schparky:
- drop `uid` from user table
- add `uuid` to user table
- sync `uuid` with [idp-id-broker](https://github.com/silinternational/idp-id-broker)
- add `uuid` to response body of `GET /user/me`

### DIFF
--- a/application/common/models/User.php
+++ b/application/common/models/User.php
@@ -48,10 +48,6 @@ class User extends UserBase implements IdentityInterface
         return ArrayHelper::merge(
             [
                 [
-                    ['uid'], 'default', 'value' => Utils::generateRandomString(),
-                ],
-
-                [
                     ['created'], 'default', 'value' => Utils::getDatetime(),
                 ],
 
@@ -70,6 +66,7 @@ class User extends UserBase implements IdentityInterface
     public function fields()
     {
         return [
+            'uuid',
             'first_name',
             'last_name',
             'idp_username',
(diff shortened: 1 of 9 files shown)


application/common/models/User.php
```diff
@@ -142,6 +139,7 @@ public static function findOrCreate($username = null, $email = null, $employeeId
         $user = self::findOne(['employee_id' => $personnelUser->employeeId]);
         if ( ! $user) {
             $user = new User();
+            $user->uuid = $personnelUser->uuid;
```
wcjr: will `uuid` be a strict API for "personnel" systems?  I ask because I was thinking the pattern used for these personnel systems was an adapter.  If so, the interface would need to identify this I would think.  For example, let's say we adapt to a personnel system that uses `id` instead of `uuid`, would this be a concern?

Schparky: That's a very good question. I've been assuming that there is only ever one personnel system (id-broker) because Id Sync handles that abstraction. If that's not the case, I think there's also a problem with other features like "hide" (Do Not Disclose). @fillup what are your thoughts?

wcjr: yeah, I think that's right, we'd never "adapt" to something other than broker here...that makes sense.

Schparky: So I wonder if we still need the `Interface`. Is this an opportunity to remove code, or is it a case of "if it's not broken don't fix it"?

wcjr: I would remove it if I knew we needed to constrain ourselves to the broker.

PR #144, opened 2019-01-07
purge old, expired resets by cron job

Schparky:
Rather than a "lazy delete" of old records, we now use cron to purge old records. Also, there is now a concept of a "grace period". After a reset has expired, we'll keep it for a little while in case the user clicks on an old link. (See PR #141)

### DIFF
--- a/application/common/config/main.php
+++ b/application/common/config/main.php
@@ -236,8 +236,9 @@ return [
         'emailServiceStatus' => $emailServiceConfig,
         'emailVerification' => $emailServiceConfig,
         'reset' => [
-            'lifetimeSeconds' => 3600, // 1 hour
-            'disableDuration' => 900, // 15 minutes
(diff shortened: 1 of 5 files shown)


application/common/config/main.php
```diff
@@ -236,8 +236,9 @@
         'emailServiceStatus' => $emailServiceConfig,
         'emailVerification' => $emailServiceConfig,
         'reset' => [
-            'lifetimeSeconds' => 3600, // 1 hour
-            'disableDuration' => 900, // 15 minutes
+            'lifetimeSeconds' => 3600,  // 1 hour
+            'gracePeriod' => '-1 week', // time between expiration and deletion
```
fillup:
```suggestion
            'gracePeriod' => '-1 week', // time between expiration and deletion, relative to now (time of execution)
```

PR #151, opened 2019-02-08
Simplify configuration and add documentation

Schparky:
- removed requirement to add a `local.php`
- removed useless/obsolete configs in `docker-compose.yml`
- documented `local.php.dist`
- added documentation of optional config in `local.php.dist`
- added documentation in `README.md`


Schparky:
Thanks. I don't know, but since it's easy to omit them, we could leave it as-is, or trim the code. I can go either way.

PR #153, opened 2019-02-21
Include last login in /user/me response

Schparky:
- added `last_login_utc` to `PersonnelUser`
- updated test -- can't compare all properties since some are dynamic, just need to ensure that the correct user is returned


api.raml
```diff
@@ -165,20 +165,28 @@ types:
       last_name: string
       idp_username: string
       email: string
-      password_meta?: PasswordMeta
+      auth_type:
+        enum: [ login, reset ]
       hide:
         enum: [ yes, no ]
+      last_login: string
```
wcjr: should this be `last_login_utc`?

api.raml
```diff
@@ -165,20 +165,28 @@ types:
...
     example:
       {
+        "uuid": "55b2447f-87f7-4286-8bd1-a7478f9dae38",
         "first_name": "John",
         "last_name": "Doe",
         "idp_username": "john_doe",
         "email": "john_doe@example.com",
+        "auth_type": "login",
+        "hide": "no",
+        "last_login": "2018-09-13T19:21:57Z",
```
wcjr: same as above comment

api.raml
Schparky: I opted to drop the `_utc` because the password meta dates don't have that suffix.

wcjr: ok, consistent among api at least.

PR #156, opened 2019-02-25
Change MFA verify response

Schparky:
- `PUT /mfa/{id}/verify` responds with the verified mfa object
- Updated dependencies, including `idp-id-broker-php-client` (dev-develop) and `yii2` (2.0.16)


application/frontend/controllers/MfaController.php
```diff
@@ -112,41 +111,45 @@ public function actionDelete($mfaId)
...
      */
     public function actionVerify($mfaId)
     {
+        $messages = [
+            400 => 'Invalid code provided',
+            404 => 'MFA verify failure',
+            429 => 'MFA rate limit failure',
+        ];
+
+        $codes = [
```
Baggerone: Maybe this way of doing things was agreed on in the past, but if not ...

I'm wondering if this may not be helpful in the end.  My understanding was that error codes as timestamps were to help us programmers find the line where the error was thrown.  

I'm thinking the message will be enough to tell us what kind of error it was, and then we can just use a single code in the `catch` block.

application/composer.json
```diff
@@ -18,7 +18,7 @@
     "silinternational/php-env": "^2.1.1",
     "silinternational/yii2-json-log-targets": "^1.0.0",
     "silinternational/yii2-email-log-target": "^1.0.1",
-    "silinternational/idp-id-broker-php-client": "^2.6.0",
+    "silinternational/idp-id-broker-php-client": "dev-develop",
```
Baggerone: I'm guessing this will need to get changed before develop is merged into master.

Schparky: Yep, I'll need to be cognizant of that. Thanks.

application/frontend/controllers/MfaController.php
Schparky: Good call, thanks.

PR #159, opened 2019-03-06
Fix bug in password metadata

Schparky:
The password expiration returned from `PUT /password` did not account for a possible mfa extension. This PR removes the local copy of the password metadata and retrieves it from the id-broker every time it's needed.


Schparky:
> any api doc updates?

Nope. Just implementation changes. Thanks for the suggestion, though.

PR #161, opened 2019-03-15
add password checks

Schparky:
* Check submitted password against those given as examples in a video about secure passwords.

### DIFF
--- a/application/common/models/Password.php
+++ b/application/common/models/Password.php
@@ -221,7 +221,7 @@ class Password extends Model
     public function validateNotPublicPassword($attribute)
     {
         /*
-         * block passwords provided in https://www.youtube.com/watch?v=3LRnjzA8auk&list=PLu5OsENIeX656zXJ96FCL169WNmvnPveo
+         * block passwords provided in https://youtu.be/WTMZYuoztoM
          */
         $publicPasswords = [
             'one4amzn',
--- a/application/common/models/Password.php
+++ b/application/common/models/Password.php
@@ -221,7 +221,7 @@ class Password extends Model
     public function validateNotPublicPassword($attribute)
     {
         /*
-         * block passwords provided in https://youtu.be/WTMZYuoztoM
+         * block passwords provided in https://youtu.be/WTMZYuoztoM?list=PLu5OsENIeX656zXJ96FCL169WNmvnPveo
          */
         $publicPasswords = [
             'one4amzn',
--- a/application/common/models/Password.php
+++ b/application/common/models/Password.php
@@ -70,10 +70,6 @@ class Password extends Model
                 'skipOnError' => false,
             ],
             [
-                'password', 'validateNotPwned',
-                'skipOnError' => true,
-            ],
-            [
                 'password', 'validateNotPublicPassword',
                 'skipOnError' => false,
             ],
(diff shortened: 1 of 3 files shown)


application/common/models/Password.php
```diff
@@ -214,5 +222,64 @@ public function save()
         }
     }
 
+    public function validateNotPublicPassword($attribute)
+    {
+        /*
+         * block passwords provided in https://www.youtube.com/watch?v=3LRnjzA8auk&list=PLu5OsENIeX656zXJ96FCL169WNmvnPveo
```
forevermatt: Minor issue, but it's probably worth wrapping just before the URL to avoid going over 120 chars on that line.

application/common/models/Password.php
```diff
@@ -214,5 +222,64 @@ public function save()
...
+        if (($httpStatusCode !== 200) || empty($results)) {
+            /*
+             * status will always be 200 if the service is working correctly
+             */
+            return;
+        }
+
+        $hashSuffix = preg_quote(substr($hash, 5));
+        $pattern = '/' . $hashSuffix . ':([0-9]+)/ism';
+        if (preg_match($pattern, $results, $matches) == 1) {
```
forevermatt: The `$matches` array hasn't been defined yet when you do this pass-by-reference. I think it will still work, but it seems worth instantiating it first (e.g. `$matches = [];`).

application/common/models/Password.php
```diff
@@ -214,5 +222,64 @@ public function save()
...
+     *
+     * @param $attribute
+     */
+    public function validateNotPwned($attribute)
+    {
+        $hash = sha1($this->$attribute);
+        $hashPrefix = substr($hash, 0, 5);
+        $url = 'https://api.pwnedpasswords.com/range/' . $hashPrefix;
+
+        $ch = curl_init();
```
forevermatt: Out of curiosity, is there a specific reason we're using `curl` rather than something like Guzzle?

It would also be good to specify a custom User-Agent string (per [the API's docs](https://haveibeenpwned.com/API/v2)), since the acceptable use policy says to the User-Agent should accurately describes the consumer of the API.

application/common/models/Password.php
```diff
@@ -214,5 +222,64 @@ public function save()
...
+     *
+     * Inspired by https://github.com/PayFast/laravel-validate-pwned-password
+     *
+     * @param $attribute
+     */
+    public function validateNotPwned($attribute)
+    {
+        $hash = sha1($this->$attribute);
+        $hashPrefix = substr($hash, 0, 5);
+        $url = 'https://api.pwnedpasswords.com/range/' . $hashPrefix;
```
forevermatt: It would probably be a good idea to [specify what version of the API we want](https://haveibeenpwned.com/API/v2#APIVersion), so that future (possibly breaking?) changes in new versions won't affect us until we've had time to refactor to support the new API version.

application/common/models/Password.php
```diff
@@ -214,5 +222,64 @@ public function save()
...
+             * status will always be 200 if the service is working correctly
+             */
+            return;
+        }
+
+        $hashSuffix = preg_quote(substr($hash, 5));
+        $pattern = '/' . $hashSuffix . ':([0-9]+)/ism';
+        if (preg_match($pattern, $results, $matches) == 1) {
+            $breachCount = $matches[1];
+            $msg = 'This password is not secure. It has been revealed in {breachCount} ' .
```
forevermatt: Slight correction: I believe the number is how many times it appeared it data breaches, not how many data breaches it appeared in. So a password that appeared 100 times in one data breach and fifty times in another data breach would have ":150" after it, not ":2".

All that to say I recommend changing... 
`revealed in {breachCount} password breaches`
... to... 
`revealed {breachCount} times in password breaches`

application/common/models/Password.php
Schparky: Regarding `curl` vs Guzzle, I don't have a preference or a good understanding of the pros/cons to make that call. The reference library used `curl` so I didn't mess with it.

I stand guilty of not thoroughly reading the API docs. Thanks for taking time to cover me.

application/common/models/Password.php
wcjr: could also look for a shortened url...and I think this video id has changed, just a moment...

wcjr: se this `https://youtu.be/WTMZYuoztoM`

wcjr:
```suggestion
         * block passwords provided in https://youtu.be/WTMZYuoztoM
```

application/common/models/Password.php
```diff
@@ -225,7 +221,7 @@ public function save()
     public function validateNotPublicPassword($attribute)
     {
         /*
-         * block passwords provided in https://www.youtube.com/watch?v=3LRnjzA8auk&list=PLu5OsENIeX656zXJ96FCL169WNmvnPveo
+         * block passwords provided in https://youtu.be/WTMZYuoztoM
```
forevermatt: If you want to keep the context of the YouTube playlist that video is in, you can use this URL:
https://youtu.be/WTMZYuoztoM?list=PLu5OsENIeX656zXJ96FCL169WNmvnPveo

If that doesn't matter, then the URL from @wcjr works great.

wcjr: the playlist is less likely to change, probably the better option

PR #162, opened 2019-03-19
use the haveibeenpwned.com API to test new passwords

Schparky:
Found a HIBP PHP library that is a better fit. Uses Guzzle and implements the API more fully. Doesn't lock to a specific API version, though.

### DIFF
--- a/application/common/models/Password.php
+++ b/application/common/models/Password.php
@@ -249,7 +249,8 @@ class Password extends Model
     }
 
     /**
-     * @param string $attribute
+     * @param string $attribute The name of the attribute being validated, typically
+     * 'password'.
      */
     public function validateNotBeenPwned($attribute)
     {
--- a/application/common/models/Password.php
+++ b/application/common/models/Password.php
@@ -3,6 +3,8 @@ namespace common\models;
 
 use common\helpers\ZxcvbnPasswordValidator;
 use common\components\passwordStore\PasswordReuseException;
+use GuzzleHttp\Exception\GuzzleException;
+use Icawebdesign\Hibp\Password\PwnedPassword;
 use yii\base\Model;
 use yii\web\BadRequestHttpException;
 use yii\web\ConflictHttpException;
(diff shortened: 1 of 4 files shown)


application/common/models/Password.php
```diff
@@ -241,4 +247,28 @@ public function validateNotPublicPassword($attribute)
             }
         }
     }
+
+    /**
+     * @param string $attribute
+     */
+    public function validateNotBeenPwned($attribute)
```
wcjr: what is `$attribute` here?  Is it a password?

Schparky: This is a Yii2 validator, so the required parameter is the name of the attribute being validated. In this case, `$attribute == 'password'`

Schparky:
> I'd suggest creating an issue in their repo for versioning the api if possible.

Unfortunately, the source is in a repository on a privately-hosted site that uses Gitlab Community Edition. I sent an email, but haven't heard anything back. Depending on how important this is, I could fork it. I did make sure to catch Guzzle exceptions and not re-throw them, so even if the API changes the end user experience should not break.

application/common/models/Password.php
wcjr: gotcha, that does ring a bell, thanks.

wcjr:
I can't predict how important at this point a versioned api will be, this will come down to a judgment call of how much we trust this author.  Your call man, you have my support either way.

forevermatt:
This looks good to me. I didn't realize it ahead of time, but sending a password (even lightly hashed) to someone else's library makes me slightly nervous. However, though his repo has very few downloads / stars, looking through the PwnedPasswords sections of his code it all seems fine at the moment.

@Schparky I'm a little confused by your comment about the code being in a privately hosted site... it looks like [the Packagist listing for it](https://packagist.org/packages/icawebdesign/hibp-php) points to a simple [GitHub repo](https://github.com/icawebdesign/hibp-php).

And, after looking some more at his code (and the API docs), it looks like I was mistaken about the API versioning issue. HIBP's main API is versioned, but the PwnedPasswords API doesn't seem to be versioned, so it's something of a non-issue apparently.

Schparky:
> @Schparky I'm a little confused by your comment about the code being in a privately hosted site... it looks like [the Packagist listing for it](https://packagist.org/packages/icawebdesign/hibp-php) points to a simple [GitHub repo](https://github.com/icawebdesign/hibp-php).

Wierd, I can only assume that he moved it since yesterday. I did send an email asking if he had it on a public site as well.

PR #163, opened 2019-03-21
Set password rule defaults

Schparky:
- Set `minLength` to 10, `maxLength` to 255, and `minScore` to 3.
- Changed `Yii::params` structure and front end `config` to remove regex patterns.


wcjr:
@fillup @forevermatt this was a change to the general thinking of these password rules.  We used to have bunches that could be turned on/off but we've decided to simply supply these 3 to every instance with defaults that ensure a minimal amount of security/integrity with the option to override them as an exception.  We need to go ahead and make these available for UAT testing but want to ensure you're aware of this since it is such a fundamental change.

application/common/config/main.php
```diff
@@ -48,10 +48,11 @@
 $idBrokerConfig['validIpRanges'] = Env::getArray('ID_BROKER_validIpRanges');
 
 $zxcvbnApiBaseUrl = Env::get('ZXCVBN_API_BASEURL');
-$passwordRules = Env::getArrayFromPrefix('PASSWORD_RULE_');
-$passwordRules['minLength'] = $passwordRules['minLength'] ?? null;
-$passwordRules['maxLength'] = $passwordRules['maxLength'] ?? null;
-$passwordRules['minScore'] = $passwordRules['minScore'] ?? null;
+
+$passwordRulesEnv = Env::getArrayFromPrefix('PASSWORD_RULE_');
+$passwordRules['minLength'] = $passwordRulesEnv['minLength'] ?? 10;
```
forevermatt: It looks like `$passwordRules` is never initialized. That still works in PHP, but it might be worth changing...
```php
$passwordRules['minLength'] = $passwordRulesEnv['minLength'] ?? 10;
$passwordRules['maxLength'] = $passwordRulesEnv['maxLength'] ?? 255;
$passwordRules['minScore'] = $passwordRulesEnv['minScore'] ?? 3;
```
... to something like...
```php
$passwordRules = [
    'minLength' => $passwordRulesEnv['minLength'] ?? 10,
    'maxLength' => $passwordRulesEnv['maxLength'] ?? 255,
    'minScore' => $passwordRulesEnv['minScore'] ?? 3,
];
```
Just a suggestion.

Schparky: Thanks, I'll make this edit.

PR #170, opened 2019-04-05
Add password assess endpoint

Schparky:
This new API endpoint (`PUT /password/assess`) runs all of the password checks, but does not save the password. All checks but one run locally. Password reuse is checked by id-broker. It is assumed that the other password interfaces do not have this capability (check without save), so the password reuse check will be skipped if the PasswordStore is anything other than ID Broker.


application/frontend/controllers/PasswordController.php
```diff
@@ -74,4 +73,46 @@ public function actionUpdate()
...
+        $user = \Yii::$app->user->identity;
+
+        $testPassword = Password::create($user, $newPassword);
+
+        if (! $testPassword->validate('password')) {
+            $errors = join(', ', $testPassword->getErrors('password'));
+            \Yii::warning([
+                'action' => 'password/assess',
+                'status' => 'error',
+                'employee_id' => $user->employee_id,
```
fillup:
```suggestion
                'email' => $user->email,
```
Can we change this to email to make searching logs easier?

PR #179, opened 2019-05-09
use memcache for session storage

Schparky:
Updated verison of PR #69. Merged feature/memcache with current develop branch, tested with [idp-profile-ui](https://github.com/silinternational/idp-profile-ui), and added `ext-memcache` to `composer.json`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM silintl/php7:7.2
 MAINTAINER Phillip Shipley <phillip_shipley@sil.org>
 
-ENV REFRESHED_AT 2016-04-22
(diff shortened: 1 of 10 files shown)


Dockerfile
```diff
@@ -3,6 +3,10 @@ MAINTAINER Phillip Shipley <phillip_shipley@sil.org>
 
 ENV REFRESHED_AT 2016-04-22
```
fillup: Go ahead and update this date to today as well

fillup:
```suggestion
ENV REFRESHED_AT 2019-05-09
```

PR #184, opened 2019-06-11
Allow partial success in Multiple component

Schparky:
Attempt to set password in every passwordStore component even if one fails. Provide messaging to indicate which failed and which succeeded.

### DIFF
--- a/application/common/components/passwordStore/Multiple.php
+++ b/application/common/components/passwordStore/Multiple.php
@@ -60,11 +60,9 @@ class Multiple extends Component implements PasswordStoreInterface
     }
     
     /**
-     * See if all of the password stores seem to be available/responding, and if
-     * so set the user's password in all of the defined password stores. If any
-     * of the password stores fail the "pre-check", this will not attempt to set
-     * the user's password on any of them, instead throwing a
-     * PasswordStoreException.
+     * Set the user's password in all of the defined password stores. If any
+     * of the password stores fail, a PasswordStoreException will be thrown
+     * with a message detailing which ones succeeded and which ones failed.
      *
      * NOTE: If successful, this will return the UserPasswordMeta returned by
      *       the first password store defined in its list.
--- a/application/common/components/passwordStore/Multiple.php
+++ b/application/common/components/passwordStore/Multiple.php
@@ -21,36 +21,6 @@ class Multiple extends Component implements PasswordStoreInterface
 
     public $displayName = 'Multiple';
 
-    /**
-     * See if all the password store backends are available.
-     *
-     * @param string $employeeId The Employee ID to use to see if each password
-     *     store is available.
-     * @param string $taskDescription A short description of what is about to be
-     *     attempted (e.g. 'set the password') if all backends are available.
-     * @throws PasswordStoreException
-     */
-    protected function assertAllBackendsAreAvailable(
-        $employeeId,
-        $taskDescription
-    ) {
-        foreach ($this->passwordStores as $passwordStore) {
-            try {
-                $passwordStore->getMeta($employeeId);
-            } catch (Exception $e) {
-                throw new PasswordStoreException(sprintf(
-                    'Did not attempt to %s because not all of the backends are '
-                    . 'available. The %s password store gave this error when '
-                    . 'asked for the specified user (%s): %s',
-                    $taskDescription,
-                    \get_class($passwordStore),
-                    var_export($employeeId, true),
-                    $e->getMessage()
-                ), 1498163919, $e);
-            }
-        }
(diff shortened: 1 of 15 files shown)


application/common/components/passwordStore/Multiple.php
```diff
@@ -88,48 +60,70 @@ public function getMeta($employeeId): UserPasswordMeta
...
+                    'Multiple.SetFailed {errors}',
+                    [
+                        'errors' => implode(', ', $errors),
+                    ]
+                );
             }
+            throw new PasswordStoreException($errorMessage, 1498162884);
         }
+
         return $responses[0];
```
forevermatt: This isn't a change made by this PR, but is there anything checking that there is at least one password store?

application/common/components/passwordStore/Multiple.php
```diff
@@ -88,48 +60,70 @@ public function getMeta($employeeId): UserPasswordMeta
...
+                ]);
+                $errors[] = $passwordStore->getDisplayName();
+            }
+        }
+
+        if (count($errors) > 0) {
+            if (count($successes) > 0) {
+                $errorMessage = \Yii::t(
+                    'app',
+                    'Multiple.SetPartialSuccess {successes} {errors}',
```
forevermatt: You may want to put a more visible delimiter between the successes and errors than just a space (especially since names can include a space), possibly with some sort of label of which list is which.

For example, `Multiple.SetPartialSuccess Google, ID Broker LDAP` doesn't seem to clearly indicate that "Google" and "ID Broker" succeeded and "LDAP" failed.

application/common/components/passwordStore/Multiple.php
```diff
@@ -165,4 +159,12 @@ public function assess($employeeId, $password)
         }
         return true;
     }
+
+    /**
+     * {@inheritdoc}
```
forevermatt: This is closer to a preference than an objective bit of feedback, but I've found that `{@inheritdoc}` doesn't end up adding any significant value after all. PhpStorm seems to show the overridden function's documentation on Ctrl-Q even without that. However, you may have some clear benefit in mind. Either way is okay with me.

application/common/models/Password.php
```diff
@@ -185,9 +186,12 @@ public function save()
             /*
              * Throw exception based on exception type
              */
-            if ($e instanceof  PasswordReuseException) {
+            if ($e instanceof PasswordReuseException) {
                 \Yii::warning($log);
                 throw new ConflictHttpException(\Yii::t('app', 'Password.PasswordReuse'), 1469194882);
+            } elseif ($e instanceof PasswordStoreException) {
+                \Yii::warning($log);
+                throw new BadRequestHttpException($e->getMessage());
```
forevermatt: Is a password store exception more likely to be due to a bad request or a server error? It seems like it could be either, but I don't know how to then decide which type of exception to throw here even if we wanted to take that into account.

application/common/components/passwordStore/Multiple.php
forevermatt: Ohhh... this isn't actually the error message. Wow, I didn't realize that just from reading this code.

application/common/components/passwordStore/Multiple.php
Schparky: Yes, in `init()`
```php
        if (empty($this->passwordStoresConfig)) {
            throw new InvalidArgumentException(
                'You must provide config for at least one password store.',
                1498162679
            );
        }
```

application/common/components/passwordStore/Multiple.php
Schparky: Yeah, I had been omitting the interpolations in the translation key string, but this one choked the unit test because it was looking at the content of the errors list. If I could figure out how to initialize Yii in behat, I could probably change it to simply `'Multiple.SetPartialSuccess'`

application/common/components/passwordStore/Multiple.php
Schparky: Not much thought, other than that I like the visual separation between methods.

application/common/models/Password.php
Schparky: Good question. I would have to analyze all of the different passwordStore components' error possibilities. In the end, I don't think it matters because the UI handles all errors the same.

application/common/components/passwordStore/Multiple.php
forevermatt: Okay, great. Thanks.

application/common/models/Password.php
forevermatt: Ahh, okay.

PR #195, opened 2019-06-26
Google Password Sync

Schparky:
### Changed
- Attempt to set password in every passwordStore component even if one fails.
### Added
- Add configurable option in Google passwordStore to look up user by employee id


application/common/components/passwordStore/Google.php
```diff
@@ -247,4 +267,49 @@ public function assess($employeeId, $password)
...
+        $usersResource = $this->getUsersResource();
+        $response = $usersResource->listUsers([
+                'domain' => $this->searchDomain,
+                'maxResults' => 2,
+                'query' => 'externalId=' . $employeeId,
+        ]);
+
+        if ($response === null || count($response['users']) === 0) {
+            \Yii::warning([
+                'action' => 'getEmailFromGoogle',
```
Baggerone: Does it matter that 'getEmailFromGoogle' doesn't match the function name ('getUserByEmployeeId')  - here and on line 304.

Schparky: Good find. That's leftover from some refactoring.

PR #207, opened 2020-05-04
no log delay

Schparky:
Configure Yii logging to immediately output log messages, rather than cache them up. Prevents messages from being lost if the container crashes (e.g. out of memory).

### DIFF
--- a/application/common/config/main.php
+++ b/application/common/config/main.php
@@ -101,6 +101,7 @@ return [
                     'logVars' => [],
                     'categories' => ['application'],
                     'prefix' => $logPrefix,
+                    'exportInterval' => 1,
                 ],
                 [
                     'class' => JsonStreamTarget::class,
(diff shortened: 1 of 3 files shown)


application/frontend/config/main.php
```diff
@@ -91,6 +91,7 @@
             ]
         ],
         'urlManager' => [
+            'cache' => null,
```
mtompset, line 94: "Set this property to false or null if you do not want to cache the URL rules."
https://www.yiiframework.com/doc/api/2.0/yii-web-urlmanager

application/common/config/main.php
```diff
@@ -101,6 +101,7 @@
                     'logVars' => [],
                     'categories' => ['application'],
                     'prefix' => $logPrefix,
+                    'exportInterval' => 1,
```
mtompset, line 104: "How many messages should be accumulated before they are exported. Defaults to 1000."
https://www.yiiframework.com/doc/api/2.0/yii-log-target#$exportInterval-detail

application/frontend/config/main.php
forevermatt: What does not-caching URL rules gain us?

Schparky: I believe this is specifying the default, which we have been using implicitly. Specifying it here silences a warning message.

PR #214, opened 2020-12-23
patch to help recover from an expired login session

Schparky:
A user will see an error page from the API if they sit on the IDP login prompt long enough for the session to expire, and then proceed with the login attempt. This patch does a redirect back to a profile UI page, which should restart the login process.

image

### DIFF
--- a/application/frontend/controllers/AuthController.php
+++ b/application/frontend/controllers/AuthController.php
@@ -46,8 +46,9 @@ class AuthController extends BaseRestController
 
     public function actionLogin()
     {
+        $afterLogin = $this->getAfterLoginUrl($this->getReturnTo());
+
         if ( ! \Yii::$app->user->isGuest) {
-            $afterLogin = $this->getAfterLoginUrl($this->getReturnTo());
             return $this->redirect($afterLogin);
         }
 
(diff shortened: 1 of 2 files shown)


longrunningprocess (commented):
I think I'll leave this one to @forevermatt .  These changes look goo d but some of this PHP can be a little brittle (imo) and moving L50 before the condition _might_ cause issues if assumptions in state were made...sorry, wish I could just confidently dismiss it as non-intrusive.

application/frontend/controllers/AuthController.php
```diff
@@ -63,7 +64,12 @@ public function actionLogin()
             try {
                 $clientId = Utils::getClientIdOrFail();
             } catch (\Exception $e) {
-                throw new BadRequestHttpException(\Yii::t('app', 'Auth.MissingClientID'), 1545316879);
+                \Yii::warning(\Yii::t('app', 'Auth.MissingClientID'));
+
+                // This condition happens if a user sits on the IDP login prompt long
+                // enough for the session to expire. As a workaround, redirect back to
+                // a profile UI page, which should restart the login process
+                return $this->redirect($afterLogin);
```
forevermatt, line 72: I believe these changes will work, but there are enough possible paths through this code that I'm not sure I've considered them all.

Since the scenario we're trying to improve here is someone leaving their browser on the login page for over 30 minutes and then trying to log in, I'm wondering about ignoring the Return-To URL and simply sending them just to the plain UI_URL. Thoughts?

I think the question comes down to whether we trust the Return-To URL more/less depending on whether the user has/hasn't already authenticated. We already (currently) ensure that it begins with the UI_URL, so I don't know whether it matters... just trying to think through the repercussions of the change.

Schparky: I like that suggestion. Thanks.

forevermatt (approved):

PR #224, opened 2021-05-25
dependency update

Schparky:


Dockerfile
```diff
@@ -1,18 +1,54 @@
...
-
-ENV REFRESHED_AT 2020-04-07
+FROM silintl/php7-apache:7.4.19
+LABEL maintainer="Phillip Shipley <phillip_shipley@sil.org>"
 
 RUN apt-get update -y && \
-    apt-get install -y php-memcache && \
-    apt-get clean
+    apt-get install -y \
+        git \
```
forevermatt, line 6: Why is `git` installed?

forevermatt (approved):
Approved, with one question

Dockerfile
Schparky: Good question. I ran into this error from composer: "git was not found in your PATH, skipping source download" and that was the only way I could find to get past it. Have you seen that before?

PR #225, opened 2021-05-26
Release 5.5.0

Schparky:
### Added
- In Google adapter, search by email as well as by external ID


Baggerone (approved):

forevermatt (approved):

application/common/components/passwordStore/Google.php
```diff
@@ -178,16 +170,40 @@ public function getMeta($employeeId): UserPasswordMeta
...
-     * @return Google_Service_Directory_User The user record from Google.
+     * @return Google_Service_Directory_User|null The user record from Google.
+     * @throws UserNotFoundException if email not defined or user not found in Google
      */
-    protected function getUser($employeeId)
+    protected function getUser(string $employeeId): ?Google_Service_Directory_User
     {
-        if ($this->findByExternalId === true) {
+        $email = $this->getEmailFromLocalStore($employeeId);
+        $user = $this->getUserByEmail($email);
```
forevermatt, line 179: It looks like `getUserByEmail()` can't return `null`... is that right? It seems to throw a UserNotFoundException.

In that case, I think we don't need the following if-block. What impact would that have on the intended behavior of the system?

application/run.sh
```diff
@@ -13,7 +13,9 @@ runny /data/yii migrate --interactive=0
 
 if [[ $APP_ENV == "dev" ]]; then
     export XDEBUG_CONFIG="remote_enable=1 remote_host="$REMOTE_DEBUG_IP
-    apt-get install php-xdebug
+    apt-get update && apt-get install -y build-essential
```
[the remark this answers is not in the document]
Schparky: It's only installed if APP_ENV is "dev" so that it can be used for remote debugging.

application/common/components/passwordStore/Google.php
Schparky: Good observation. Looks like my brain is still in Go mode...not catching exceptions.

Schparky: Addressed this in 5.5.1

PR #229, opened 2021-06-07
Upgrade to PHP 7.4 (again)

forevermatt:
### Changed
- Upgrade PHP to 7.4 again (but stayed with Ubuntu base O/S)
- Upgrade Ubuntu to 21.04
- Stop including syslog and Logentries-related files

### Fixed
- Upgrade Composer to 2
- Switch from fxp's composer-asset-plugin to asset-packagist

---

I'm not sure whether to treat this as a breaking change. Thoughts?

P.S. The failing Codeship builds seem to be related to [this issue](https://stackoverflow.com/questions/66319610/gpg-error-in-ubuntu-21-04-after-second-apt-get-update-during-docker-build)


forevermatt:
Response from Codeship support:
> Currently it is not possible to change the Docker version from your side. This is something that is in the Engineering Teams backlog, but I cannot provide any estimation on when this change may be done.
> 
> I will also share this feedback with our Product Team so they are aware of that issue.

I do appreciate how quickly they responded, but it sounds like we'll need to find some other way around this problem in the meantime...

Schparky:
I don't see a breaking change. Either I'm missing something important, or have a different understanding about what would be a BC.

Schparky:
Also, I don't see the connection between the SO page and the error. I see this in the Codeship logs:

> /data/run-tests-api.sh: line 18: /data/vendor/bin/codecept: No such file or directory

forevermatt:
> I don't see a breaking change. Either I'm missing something important, or have a different understanding about what would be a BC.

I was wondering whether removing rsyslog and the Logentries helper script (done in the new parent image) would qualify, since anyone using this image might be depending on rsyslog for some reason. That's probably a stretch, so we could just go with a patch release.

forevermatt:
Closing this due to further changes made that invalidate the branch name

PR #249, opened 2022-12-14
Multi webauthn

Baggerone:


Schparky (approved):
Assuming this isn't a code review so much as an approval process to move this one step closer to release. As such, I will approve, knowing that you have a timeline in mind for synchronizing these changes with the other pieces of this enhancement.

Schparky:
I just noticed that there are API changes, but no edits to the `api.raml` file.

Schparky:
> Assuming this isn't a code review so much as an approval process to move this one step closer to release. As such, I will approve, knowing that you have a timeline in mind for synchronizing these changes with the other pieces of this enhancement.

I guess this isn't a develop branch like on idp-id-broker. I'll take a more thorough look at this one.

Schparky (approved):

application/frontend/controllers/MfaController.php
```diff
@@ -140,17 +140,40 @@ public function actionDelete($mfaId)
...
+    }
 
+    /**
+     * @return array
+     * @throws HttpException
+     */
+    private static function getVerifyValue() {
         $value = \Yii::$app->request->getBodyParam('value');
         if ($value === null) {
             throw new BadRequestHttpException(\Yii::t('app', 'Mfa.MissingValue'));
```
forevermatt, line 179: Now that this is in a separate private method (rather than in the controller action), it would probably be better to simply let this `getVerifyValue()` method return `null`, and leave the calling method (the actual controller action that's handling the request) throw the `BadRequestHttpException` if the value is `null`.
```suggestion
        if ($value === null) {
            return null;
```

application/frontend/controllers/MfaController.php
```diff
@@ -161,6 +184,24 @@ public function actionVerify($mfaId)
...
+     */
+    public function actionVerify($mfaId)
+    {
+        $messages = [
+            400 => \Yii::t('app', 'Mfa.InvalidCode'),
+            404 => \Yii::t('app', 'Mfa.VerifyFailure'),
+            429 => \Yii::t('app', 'Mfa.RateLimitFailure'),
+        ];
+
+        $value = self::getVerifyValue();
```
forevermatt, line 203: Related to my comment above about returning `null` from the `getVerifyValue()` method:
```suggestion
        $value = self::getVerifyValue();
        if ($value === null) {
            throw new BadRequestHttpException(\Yii::t('app', 'Mfa.MissingValue'));
        }
```

application/frontend/controllers/MfaController.php
```diff
@@ -185,6 +226,46 @@ public function actionVerify($mfaId)
...
+     */
+    public function actionVerifyregistration($mfaId)
+    {
+        $messages = [
+            400 => \Yii::t('app', 'Mfa.InvalidCode'),
+            404 => \Yii::t('app', 'Mfa.VerifyFailure'),
+            429 => \Yii::t('app', 'Mfa.RateLimitFailure'),
+        ];
+
+        $value = self::getVerifyValue();
```
forevermatt, line 242: Also related to my comment above about returning `null` from the `getVerifyValue()` method:
```suggestion
        $value = self::getVerifyValue();
        if ($value === null) {
            throw new BadRequestHttpException(\Yii::t('app', 'Mfa.MissingValue'));
        }
```

application/frontend/controllers/MfaController.php
```diff
@@ -140,17 +140,40 @@ public function actionDelete($mfaId)
...
+
+            /*
+             * Other status codes will result in a 500 response
+             */
+            throw $e;
+        }
+    }
 
+    /**
+     * @return array
```
forevermatt, line 173: See comment below.
```suggestion
     * @return array|null
```

application/tests/api/MfaCest.php
```diff
@@ -59,6 +59,23 @@ public function test31(ApiTester $I)
...
+        $I->seeResponseCodeIs(401);
+    }
+
+    public function test34(ApiTester $I)
+    {
+        $I->wantTo('check response when making PUT request to mfa/{id}/webauthn/{webauthn_id} for a user'
+            . ' with auth_type=reset');
+        $I->haveHttpHeader('Authorization', 'Bearer user5');
+        $I->sendPUT('/mfa/5/webauthn/6');
+        $I->seeResponseCodeIs(403);
```
forevermatt, line 76: I think the RAML will need to be updated for that endpoint. It doesn't seem to include `403` as a possible response code for `PUT /mfa/{id}/webauthn/{webauthn_id}`.

forevermatt (changes requested):

application/frontend/config/main.php
```diff
@@ -156,14 +156,19 @@
...
-                'PUT /mfa/<mfaId:(\d+)>/verify'     => 'mfa/verify',
-                'OPTIONS /mfa'                      => 'mfa/options',
-                'OPTIONS /mfa/<mfaId:(\d+)>'        => 'mfa/options',
-                'OPTIONS /mfa/<mfaId:(\d+)>/verify' => 'mfa/options',
+                'GET /mfa'                                               => 'mfa/index',
+                'POST /mfa'                                              => 'mfa/create',
+                'PUT /mfa/<mfaId:(\d+)>'                                 => 'mfa/update',
+                'PUT /mfa/<mfaId:(\d+)>/webauthn/<webauthnId:(\d+)>'     => 'mfa/updatewebauthn',
+                'DELETE /mfa/<mfaId:(\d+)>'                              => 'mfa/delete',
+                'DELETE /mfa/<mfaId:(\d+)>/webauthn/<webauthnId:(\d+)>'  => 'mfa/deletewebauthn',
```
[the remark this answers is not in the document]
Baggerone: Thanks for the tip/reminder.

application/frontend/controllers/MfaController.php
Baggerone: I don't mind changing it, but I'm confused why you would want an `if` block repeated in more than one calling method when you can have it just once in the called method.

application/frontend/controllers/MfaController.php
```diff
@@ -215,4 +301,41 @@ public function actionUpdate($mfaId)
...
+     * @param $mfaId
+     * @param $webauthnId
+     * @throws NotFoundHttpException
+     * @throws ServiceException
+     */
+    public function actionUpdateWebauthn($mfaId, $webauthnId)
+    {
+        $label = \Yii::$app->request->getBodyParam('label');
+        if ($label === null) {
+            \Yii::$app->response->statusCode = 204;
```
forevermatt, line 315: I just noticed that this returns a `204` if no label was provided, but the RAML looks like it says it will return a missing or blank label value. Is that accurate? Which do we want it to do?

Baggerone: Good catch. I'm fixing this, although I don't see in the RAML where it says it will return a a missing or blank label value.

forevermatt (approved):

application/frontend/controllers/MfaController.php
forevermatt: You made a good point. As long as the HTTP exception is still in the controller (which it is in both of these options), it's still in a good location. Thanks for pushing back on that... you're right.

PR #257, opened 2023-05-15
Move session storage from memcache to database

Baggerone:


application/frontend/config/main.php
```diff
@@ -47,20 +38,18 @@
...
+            'class' => 'yii\web\DbSession',
+             'db' => [
+                 'class' => 'yii\db\Connection',
+                 'dsn' => sprintf('mysql:host=%s;dbname=%s', $mysqlHost, $mysqlDatabase),
+                 'username' => $mysqlUser,
+                 'password' => $mysqlPassword,
+                 'charset' => 'utf8',
+                 'emulatePrepare' => false,
+                 'tablePrefix' => '',
+             ],
```
forevermatt, line 51: Looks like these ended up with an extra space of indentation:
```suggestion
            'db' => [
                'class' => 'yii\db\Connection',
                'dsn' => sprintf('mysql:host=%s;dbname=%s', $mysqlHost, $mysqlDatabase),
                'username' => $mysqlUser,
                'password' => $mysqlPassword,
                'charset' => 'utf8',
                'emulatePrepare' => false,
                'tablePrefix' => '',
            ],
```

application/requirements.php
```diff
@@ -78,13 +78,6 @@
         'memo' => 'Required for PostgreSQL database.',
     ),
     // Cache :
-    array(
-        'name' => 'Memcache extension',
-        'mandatory' => false,
-        'condition' => extension_loaded('memcache') || extension_loaded('memcached'),
-        'by' => '<a href="http://www.yiiframework.com/doc-2.0/yii-caching-memcache.html">MemCache</a>',
-        'memo' => extension_loaded('memcached') ? 'To use memcached set <a href="http://www.yiiframework.com/doc-2.0/yii-caching-memcache.html#$useMemcached-detail">MemCache::useMemcached</a> to <code>true</code>.' : ''
-    ),
```
forevermatt, line 87: I think this could be left in, unless it caused problems. It's not code we wrote... I think it's just for providing info (if we manually run it) about which of the features that Yii supports we could use, based on whether the feature's requirements are met.

forevermatt (approved):
Approved, with minor suggestions

application/console/migrations/m230515_162100_create_session_table.php
```diff
@@ -0,0 +1,28 @@
...
+    {
+        /*
+         * UserTable
+         */
+        $this->createTable(
+            '{{session}}',
+            [
+                'id' => 'char(40) primary key not null',
+                'expire' => 'int(11) null',
+                'data' => 'binary not null',
```
briskt, line 18:
```suggestion
                'data' => 'blob not null',
```
According to the [DbSession docs](https://www.yiiframework.com/doc/api/2.0/yii-web-dbsession#$sessionTable-detail) this should be a `LONGBLOB`. I'm not sure we ever store that much data in the session, but `binary` is much shorter (255 bytes).

application/console/migrations/m230515_162100_create_session_table.php
```diff
@@ -0,0 +1,28 @@
...
+            ],
+            "ENGINE=InnoDB DEFAULT CHARSET=utf8"
+        );
```
briskt, line 21:
```suggestion
        );
        $this->createIndex('idx_expire', 'session', 'expire', false);
```
The docs recommend adding an index in production. We may not scale large enough to notice the difference, but I suppose it can't hurt.

application/console/migrations/m230515_162100_create_session_table.php
```diff
@@ -0,0 +1,28 @@
...
+    // Use safeUp/safeDown to run migration code within a transaction
+    public function safeUp()
+    {
+        /*
+         * UserTable
+         */
+        $this->createTable(
+            '{{session}}',
+            [
+                'id' => 'char(40) primary key not null',
```
briskt, line 16:
```suggestion
                'id' => 'char(64) primary key not null',
```
In case we decide to use a hashing algorithm other than MD5.

application/versions.txt
```diff
@@ -2,28 +2,28 @@ adldap2/adldap2                           v10.5.0            A PHP LDAP Package
 behat/behat                               v3.13.0            Scenario-oriented BDD framework for PHP
 codeception/codeception                   5.0.10             BDD-style testing framework
 codeception/module-asserts                2.0.1              Codeception module containing various assertions
-codeception/module-phpbrowser             3.0.0              Codeception module for testing web application o...
+codeception/module-phpbrowser             3.0.0              Codeception module for testing web application over HTTP
```
briskt, line 5: Not that it's important, but you can get the new text to match the old by changing the width of your terminal window.

briskt (changes requested):

PR #259, opened 2023-06-07
Fix the mock test personnel component so that the tests pass again.

Baggerone:
Two tests had been failing.


briskt (approved):
good one

jason-jackson (approved):
:+1:

PR #266, opened 2023-10-09
use YII_ENV instead of APP_ENV

briskt:
### Fixed
- Simplification: use YII_ENV in main.php

---

### Feature PR Checklist
- [x] Run `make composershow`

### DIFF
--- a/application/common/config/main.php
+++ b/application/common/config/main.php
@@ -142,7 +142,7 @@ return [
                     'assertValidIp' => $emailServiceConfig['assertValidIp'],
                     'validIpRanges' => $emailServiceConfig['validIpRanges'],
                     'enabled' => $alertsEmailEnabled,
-                    'prefix' => function($message) use (YII_ENV) {
+                    'prefix' => function($message) {
                         $prefixData = [
                             'env' => YII_ENV,
                         ];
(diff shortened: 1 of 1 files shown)


application/common/config/main.php
```diff
@@ -136,16 +135,16 @@
...
-                        'subject' => 'ALERT - ' . $idpName . ' PW [env=' . $appEnv . ']',
+                        'subject' => 'ALERT - ' . $idpName . ' PW [env=' . YII_ENV . ']',
                     ],
                     'baseUrl' => $emailServiceConfig['baseUrl'],
                     'accessToken' => $emailServiceConfig['accessToken'],
                     'assertValidIp' => $emailServiceConfig['assertValidIp'],
                     'validIpRanges' => $emailServiceConfig['validIpRanges'],
                     'enabled' => $alertsEmailEnabled,
-                    'prefix' => function($message) use ($appEnv) {
+                    'prefix' => function($message) use (YII_ENV) {
```
forevermatt, line 145: I don't think you need this in the `use` parens, since it's a global constant:
```suggestion
                    'prefix' => function($message) {
```

I think that's why the build failed:
> PHP Parse error:  syntax error, unexpected identifier "YII_ENV", expecting variable or "&" or amp in /data/common/config/main.php on line 145

forevermatt (changes requested):
See comment

application/common/config/main.php
briskt: I did that correct in id-broker, and I saw the red squiggle under the filename. But when I hovered over the scrollbar it showed a warning that I didn't need to address (Unused parameter "message"). Should have looked a little closer.

forevermatt (approved):

PR #283, opened 2024-06-05
IDP-1076: Add option to require alpha and numeric chars for PCI 4.0 compliance

jason-jackson:
### Added
- Add validation option to require alpha and numeric chars

Note: Reference for regex is [here](https://www.php.net/manual/en/regexp.reference.unicode.php).

---

### Feature PR Checklist
- [ ] Run `make composershow`

### DIFF
--- a/application/common/models/Password.php
+++ b/application/common/models/Password.php
@@ -87,7 +87,14 @@ class Password extends Model
             [
                 'password', 'validateNoBadBytes',
                 'skipOnError' => false,
-            ]
+            ],
+            [
+                'password', 'validateAlphaAndNumeric',
+                'skipOnError' => false,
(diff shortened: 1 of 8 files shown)


application/common/models/Password.php
```diff
@@ -288,4 +295,18 @@ public function validateNoBadBytes($attribute)
             $this->addError($attribute, \Yii::t('app', 'Password.ContainsBadByte'));
         }
     }
+
+    public function validateAlphaAndNumeric($attribute)
```
briskt, line 299:
```suggestion
    public function validateAlphaAndNumeric(string $attribute): void
```

briskt (approved):

PR #284, opened 2024-07-29
Release 6.8.0 -- new password rule `requireAlphaAndNumeric`

briskt:
### Added
- Added new password rule `requireAlphaAndNumeric`

### Removed
- Removed download of config-shim. It's now included in the base image silintl/php8:8.1


hobbitronics (approved):
I've only given a gloss over, but approving based on other's prior approval.

briskt:
> I've only given a gloss over, but approving based on other's prior approval.

I was hoping @jason-jackson would approve it. I'm not sure if it's fully ready to deploy.

jason-jackson (approved):
Seems a bit strange for me to review it, but here's my approval

briskt:
> Seems a bit strange for me to review it, but here's my approval

It's not a code review so much as saying that the timing is right. @forevermatt thought maybe there was an issue with allowing passwords that don't have a number. Did that get resolved?

jason-jackson:
I think everything is working as much as we can get it with the PHP regex (there seems to be an issue with at least one letter also counting as a number, but I can't fix that as that is a PHP issue).

briskt:
@jason-jackson when you say "at least one letter", I assume you're referring to characters like ` ü `? I think that's fine. It achieves the intended purpose.

PR #287, opened 2024-08-12
Release - move access_token to httpOnly cookie (7.0.0)

hobbitronics:
### Changed
- moved access_token to httpOnly cookie

### Removed
- client_id
---

### Feature PR Checklist
- [ ] Run `make composershow`


api.raml
```diff
@@ -4,14 +4,16 @@ version: 4
 protocols: [ HTTPS ]
 mediaType: application/json
 securitySchemes:
-  AuthzBearerToken:
-      type: x-{other}
+ AuthzHttpOnlyCookie:
```
briskt, line 7: Indentation changed on this line. Might be because this section was indented strange. Could you take a look and see what needs to be done to make it consistent with the rest of the file?

api.raml
```diff
@@ -236,18 +235,16 @@ types:
...
-            `ReturnToOnError` with no query string parameters.
+            On successful login or invite, a redirection to `ReturnTo`. If `invite` was provided, but it has expired,
+            this response will be a redirection to `ReturnToOnError` with no query string parameters.
           headers:
             location:
               description: redirect URL
               type: string
               example:
-                https://idp-pw.org/profile/intro?state=&token_type=Bearer&expires_utc=2018-12-20T18:41:52Z&access_token=hgEzk1kceC40FHeJm6cyfr5FgE9Tc4UH
+                https://idp-pw.org/profile/intro?state=&expires_utc=2018-12-20T18:41:52Z
```
briskt, line 245:
```suggestion
                https://idp-pw.org/profile/#
```
`expires_utc` isn't present when I try this on appsdev

api.raml
```diff
@@ -236,18 +235,16 @@ types:
...
           example:
-            https://idp-pw.org/profile/intro?state=&token_type=Bearer&expires_utc=2018-12-20T18:41:52Z&access_token=hgEzk1kceC40FHeJm6cyfr5FgE9Tc4UH
+            https://idp-pw.org/profile/intro?state=&expires_utc=2018-12-20T18:41:52Z
```
briskt, line 247: this example seems redundant and could be removed

application/frontend/controllers/AuthController.php
```diff
@@ -131,12 +125,15 @@ public function actionLogin()
 
     public function actionLogout()
     {
-        $accessToken = \Yii::$app->request->get('access_token');
+        $requestCookies = \Yii::$app->request->cookies;
+        $responseCookies = \Yii::$app->response->cookies;
```
briskt, line 129: could move this into the `if` block below

application/frontend/controllers/AuthController.php
```diff
@@ -209,20 +206,8 @@ public function getLoginSuccessRedirectUrl($state, $accessToken, $tokenExpiratio
...
-            '%sstate=%s&token_type=Bearer&expires_utc=%s&access_token=%s',
-            $joinChar,
-            Html::encode($state),
-            Utils::getIso8601($tokenExpiration),
-            $accessToken
-        );
-
-        return $url;
+
+        return $afterLogin;
```
briskt, line 210: could combine this with the previous line
```
return $this->getAfterLoginUrl($relayState);
```

application/frontend/controllers/AuthController.php
```diff
@@ -90,8 +75,17 @@ public function actionLogin()
...
+            $secure = YII_ENV != 'dev' || YII_ENV != 'test' ? true : false;
+
+            \Yii::$app->response->cookies->add(new \yii\web\Cookie([
+              'name' => 'access_token',
+              'value' => $accessToken,
+              'expire' => $user->access_token_expiration,
+              'httpOnly' => true, // Ensures the cookie is not accessible via JavaScript
+              'secure' => $secure,   // Ensures the cookie is sent only over HTTPS
+              'sameSite' => 'Lax', // Adjust as needed
+            ]));
```
briskt, line 88: this could be extracted to a separate function so any future changes are the same on both login and reset flows

application/tests/api/ResetCest.php
```diff
@@ -115,77 +127,79 @@ public function test7(ApiTester $I)
     public function test72(ApiTester $I)
     {
         $I->wantTo('check response when making authenticated PUT request to resend the verification');
-        $I->haveHttpHeader('Authorization', 'Bearer user1');
+        $I->setCookie('access_token', 'user1', [
+          'expire' => time() + 3600,  // Cookie expires in 1 hour
+          'httpOnly' => true          // Cookie is not accessible via JavaScript
+        ]);
```
briskt, line 133: if it's not difficult, this would be good to extract to a separate function

briskt (commented):
Excellent work on this, thank you!
I added a few suggestions. You could also solicit feedback from others if you want.

application/tests/api/ResetCest.php
hobbitronics: not sure why setting a var in _bootstrap.php isn't working, but apparently it's too hard for me at the moment. I was just planning on reusing the `[
          'expire' => time() + 3600,  // Cookie expires in 1 hour
          'httpOnly' => true          // Cookie is not accessible via JavaScript
        ]` part as `$cookieConfig`

briskt (approved):

hobbitronics:
> Excellent work on this, thank you! I added a few suggestions. You could also solicit feedback from others if you want.

Couldn't have done it without your help.

mtompset (approved):
Long PRs bad. Looks okay, but I'm not familiar with raml and the previous authentication methodology.

briskt:
> Long PRs bad.

Is that a judgment?

application/common/models/User.php
```diff
@@ -580,26 +580,35 @@ public function setPassword($newPassword)
...
          */
         $this->auth_type = $authType;
-        $this->access_token = Utils::getAccessTokenHash($clientId . $accessToken);
+        $this->access_token = Utils::getAccessTokenHash($accessToken);
         $this->access_token_expiration = Utils::getDatetime(
             time() + \Yii::$app->params['accessTokenLifetime']
         );
         $this->saveOrError('Unable to create access token', 1465833228);
 
+        $secure = (YII_ENV != 'dev' || YII_ENV != 'test') ? true : false;
```
jason-jackson, line 603: This is always true. Also, ternary does nothing. Possibly consider using !in_array() as well.

application/tests/api/ResetCest.php
```diff
@@ -115,77 +115,70 @@ public function test7(ApiTester $I)
...
         $I->wantTo('check response when making authenticated DELETE request to reset/id/validate');
-        $I->haveHttpHeader('Authorization', 'Bearer user1');
+        $I->setCookie('access_token', 'user1', parent::getCookieConfig());
         $I->sendDELETE('/reset/33333333333333333333333333333333/validate', ['code' => '333']);
         $I->seeResponseCodeIs(405);
     }
 
     public function test9(ApiTester $I)
     {
         $I->wantTo('check response when making multiple authenticated PUT request to validate a reset code');
```
jason-jackson, line 162: It might be clearer if this was in a for loop, but as this was how it already was, it doesn't need to be changed.

jason-jackson (changes requested):
Looks great thanks for doing this. The only issue I see is the yii env check always being true.

application/common/models/User.php
hobbitronics: maybe that's why I was having cors issues locally

hobbitronics: nice catch btw.

briskt:
Just wondering about the version number. Since the UI can continue to operate without change (send a client ID and/or token in the Authorization header and/or query string) I wonder if this isn't a breaking change after all. Not that bumping the version is necessarily a bad thing. It's simply that I enjoy the exercise of deciding the best way to proceed.

briskt:
> > Excellent work on this, thank you! I added a few suggestions. You could also solicit feedback from others if you want.
> 
> Couldn't have done it without your help.

You could have. I didn't do much.

hobbitronics:
> Just wondering about the version number. Since the UI can continue to operate without change (send a client ID and/or token in the Authorization header and/or query string) I wonder if this isn't a breaking change after all. Not that bumping the version is necessarily a bad thing. It's simply that I enjoy the exercise of deciding the best way to proceed.



> Just wondering about the version number. Since the UI can continue to operate without change (send a client ID and/or token in the Authorization header and/or query string) I wonder if this isn't a breaking change after all. Not that bumping the version is necessarily a bad thing. It's simply that I enjoy the exercise of deciding the best way to proceed.

While it technically works with the UI I did have trouble getting /reset/[id]/validate to work until a used the updated UI without the token so there could be bugs between the two. I was thinking of bumping the version because of that possibility as well as the intention of not having a token on the frontend with this change. I'm open to either minor or major though.

briskt:
> While it technically works with the UI I did have trouble getting /reset/[id]/validate to work until a used the updated UI without the token so there could be bugs between the two. I was thinking of bumping the version because of that possibility as well as the intention of not having a token on the frontend with this change. I'm open to either minor or major though.

OK, good to know. Let's proceed with the version bump then.

jason-jackson (approved):

application/tests/api/ResetCest.php
hobbitronics: I'll do that in a separate PR. Thanks

PR #292, opened 2024-10-10
Release 7.0.4 -- Push image to GHCR.io 

Praveenraj-K:
Added

Added a step to push the Docker image to GitHub Container Registry

### Changed
-

### Deprecated
-

### Removed
-

### Fixed
- 

### Security
-

---

### Feature PR Checklist
- [ ] Unit tests created or updated
- [ ] Run `make composershow`

### DIFF
--- a/.github/workflows/test-and-publish.yml
+++ b/.github/workflows/test-and-publish.yml
@@ -39,8 +39,8 @@ jobs:
         uses: docker/login-action@v3
         with:
           registry: ghcr.io
-          username: ${{ github.actor}}
-          password: ${{ secrets.GITHUB_TOKEN}}
+          username: ${{ github.actor }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test-and-publish.yml
```diff
@@ -26,20 +28,32 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        # Step to log into the DockerHub
```
briskt, line 31: In my opinion, comments that simply repeat what the code says should not be added. In this case "Step to log into the DockerHub" is basically the same as `name: Log in to Docker Hub`

briskt (approved):

.github/workflows/test-and-publish.yml
Praveenraj-K: got it

PR #293, opened 2024-10-30
Release 7.0.5 -- test on all branches, but only push image for tags

briskt:
### Added
- Also push a major-version tag 

### Fixed
- Run CI tests on all branches


hobbitronics (approved):
Looks good!

PR #304, opened 2025-01-31
Release 7.3.0 - Update to php 8.3

jason-jackson:
[ITSE-1300](https://itse.youtrack.cloud/issue/ITSE-1300)

---

### Changed
- Update to php 8.3
- Switch `adldap2/adldap2` to `directorytree/ldaprecord`

---

### PR Checklist

- [x] Run `make composershow`


briskt (approved):
Good to see that no code changes were needed.

forevermatt (dismissed):
If you would, please update composer dependencies along with this PHP version upgrade

application/common/components/passwordStore/Ldap.php
```diff
@@ -70,11 +70,8 @@ class Ldap extends Component implements PasswordStoreInterface
      */
     public $updateAttributesOnSetPassword = [];
 
-    /** @var \Adldap\Connections\Provider */
-    public $ldapProvider;
-
-    /** @var \Adldap\Adldap LDAP client*/
-    public $ldapClient;
+    /** @var \LdapRecord\Connection */
+    public Connection $ldapConnection;
```
jason-jackson, line 74: To keep lines changed at a minimum, I could keep the `$ldapClient` name instead of this if desired,

mtompset:
Do we even used LDAP at all? Rather than fixing the code, I think removing it might be cleaner. Less is more.

application/common/components/passwordStore/Ldap.php
```diff
@@ -312,9 +313,9 @@ protected function updatePassword($user, $password)
     }
 
     /**
-     * @param \Adldap\Models\Entry $user
+     * @param \LdapRecord\Models\OpenLDAP\Entry $user
```
mtompset, line 316: I prefer to put 'use' statements above to simplify this. Perhaps a use this as LDAP_Entry?

application/common/components/passwordStore/Ldap.php
```diff
@@ -288,11 +289,11 @@ protected function encodeForUnicodePwdField($password)
...
      * @param string $password
      * @throws \common\components\passwordStore\PasswordReuseException
      */
-    protected function updatePassword($user, $password)
+    protected function updatePassword(Entry $user, string $password): void
```
mtompset, line 296: Why is the PHPDoc the long path, but the parameter type shorter? Please LDAP_Entry would be more clear as to what Entry really means.

application/common/components/passwordStore/Ldap.php
```diff
@@ -312,9 +313,9 @@ protected function updatePassword($user, $password)
     }
 
     /**
-     * @param \Adldap\Models\Entry $user
+     * @param \LdapRecord\Models\OpenLDAP\Entry $user
      */
-    protected function removeAttributesAfterNewPassword($user)
+    protected function removeAttributesAfterNewPassword(Entry $user): void
```
mtompset, line 318: I like the stronger typing. I don't like using just Entry.

application/tests/codeception/config/unit.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
mtompset, line 2: Glad this PSR-12-ish blank line is a quick read. :)

application/tests/unit/common/components/LdapTest.php
```diff
@@ -47,8 +44,8 @@ public function testRemoveAttributesOnSet()
         /*
          * Get user before change to ensure presence of attributes to be removed
          */
-        /** @var \Adldap\Models\Entry $user */
-        $user = $ldap->ldapProvider->search()
+        /** @var \LdapRecord\Models\OpenLDAP\Entry $user */
+        $user = $ldap->ldapConnection->query()
```
mtompset, line 48: A double -> is ugly, in my opinion. Also, "slower" access. Having an $ldapConnection = $ldap->ldapConnection; would also be reusable below.

mtompset (commented):
I'd prefer recommended tiny clean ups before approving, though this code is functional (at least from eye-balling it).

briskt:
> Do we even used LDAP at all? Rather than fixing the code, I think removing it might be cleaner. Less is more.

~Unfortunately, yes.~ I was about to say we use it, but remembered that Wycliffe doesn't use it any longer and they were the only ones that did. Rather than remove it in this PR, I would recommend removing it first and then revisit this upgrade. But I guess you can proceed with this change and then rip it out.

briskt (changes requested):
Since we can't test that this new LDAP implementation works, I don't like the idea of pushing out a release with it.

jason-jackson:
> > Do we even used LDAP at all? Rather than fixing the code, I think removing it might be cleaner. Less is more.
> 
> ~Unfortunately, yes.~ I was about to say we use it, but remembered that Wycliffe doesn't use it any longer and they were the only ones that did. Rather than remove it in this PR, I would recommend removing it first and then revisit this upgrade. But I guess you can proceed with this change and then rip it out.

Yeah, I don't like doing both. Other option would be to upgrade PHP without any dependencies updated (I think everything still works) and then remove this.

briskt:
> > > Do we even used LDAP at all? Rather than fixing the code, I think removing it might be cleaner. Less is more.
> > 
> > 
> > ~Unfortunately, yes.~ I was about to say we use it, but remembered that Wycliffe doesn't use it any longer and they were the only ones that did. Rather than remove it in this PR, I would recommend removing it first and then revisit this upgrade. But I guess you can proceed with this change and then rip it out.
> 
> Yeah, I don't like doing both. Other option would be to upgrade PHP without any dependencies updated (I think everything still works) and then remove this.

There were a lot of dependency updates, so it's kind of a gamble what will or will not work on 8.3

mtompset:
If there are models involved, the base models may not be generating well since Yii 2.0.52. Check the scoping and data typing in the base models on constants. If it is lacking, check out MailAdmin's gii folder, and console.php in the config directory. Then generate with "psr-12" like in scripts' tables.sh, which is triggered by make baseModels.

forevermatt:
@jason-jackson Is this PR still in progress? I see that we have a separate 7.2.0 release already (so this might be 7.3.0 now), but those changes haven't been pulled into this feature branch yet.

I wonder if this should be set aside until after we finish removing LDAP support (if we can / are going to do that). Or maybe that's already your plan?

If needed, you could also do a separate dependency update first, then do the PHP-8.3-upgrade-with-another-dependency-update, to help keep each set of changes smaller. Just an idea, though I don't know if that would help.

application/common/components/passwordStore/Ldap.php
briskt: That might be preferred, but not worth changing back at this point.

application/common/components/passwordStore/Ldap.php
briskt: It's a bit risky on old code, but possibly worth it.

briskt (approved):
Thanks for patiently working through this.

briskt:
> I see that we have a separate 7.2.0 release already (so this might be 7.3.0 now)

I edited the PR title

application/tests/unit/common/components/LdapTest.php
jason-jackson: I don't think it is reusable. I think it needs to be set again after every `$ldap->connect()` call.

jason-jackson:
> @jason-jackson Is this PR still in progress? I see that we have a separate 7.2.0 release already (so this might be 7.3.0 now), but those changes haven't been pulled into this feature branch yet.
> 
> I wonder if this should be set aside until after we finish removing LDAP support (if we can / are going to do that). Or maybe that's already your plan?
> 
> If needed, you could also do a separate dependency update first, then do the PHP-8.3-upgrade-with-another-dependency-update, to help keep each set of changes smaller. Just an idea, though I don't know if that would help.

Yes it's still in progress, but should be good to look at again. I'll need a final approval from you (@forevermatt) to remove the changes requested so I can merge this (date tbd). We'll have to do a quick sync with JAARS to ensure it still works and then roll back if it doesn't.

mtompset (approved):
Are the reconnections in the same test necessary? This was just cleaning up generally though, so it's okay.

PR #324, opened 2025-07-17
correct incompatible default values in inherited function signature

briskt:
### Fixed
- In RedirectException, change the default constructor parameter values to match the parent types.
- Minor documentation fixes


hobbitronics (approved):
Nice improvements

PR #364, opened 2026-05-15
Release 10.0.0-pre.1 -- Remove the "type" property on the Reset model

briskt:
[IDP-1856](https://support.gtis.sil.org/issue/IDP-1856) Make Do Not Disclose the standard, then remove unnecessary code

---

### Removed
- Due to the ambiguity intentionally added to the password reset flow, there is no reason to keep the concept of a reset "type". All available password reset email addresses (except supervisor) are used every time instead of letting the user choose which one to use at the time of recovery. This PR doesn't fully remove the database column or the model attribute, but removes all means to change the attribute.
- Removed the `GET /reset/{uid}` and `PUT /reset/{uid}/resend` endpoints. These are no longer in use due the simpler reset flow. There is no use in retrieving the reset as there is no useful metadata available, and a re-send is done by repeating the `POST /reset` request.

---

### PR Checklist

- [ ] Run `make composershow`

### DIFF
--- a/api.raml
+++ b/api.raml
@@ -588,52 +588,6 @@ types:
     responses:
       204:
   /{resetId}:
-    get:
-      securedBy: null
-      description: Request the reset object by its id.
-      responses:
-        200:
-          body:
-            type: Reset
-    put:
-      securedBy: null
-      description: >
-        Update the reset to use a different email. If `type` is `email`,
-        then this request is for an alternate recovery Method, and
-        `methodId` is required.
-      body:
-        properties:
-          type:
-            enum: [ primary, supervisor, email ]
-          methodId?: string
-      responses:
-        200:
-          body:
-            type: Reset
-        400:
-          description: >
-            The reset `type` was not provided or is not valid, or the
-            `methodId` parameter was not provided when required.
-          body:
-            type: Error
-        404:
-          description: >
-            The reset could not be found or the referenced Method has
-            not been verified.
-          body:
-            type: Error
-    /resend:
-      put:
-        securedBy: null
-        description: Request to resend verification email.
-        responses:
-          200:
-            body:
-              type: Reset
-          404:
-            description: Reset could not be found.
-            body:
-              type: Error
     /validate:
       put:
         securedBy: null
--- a/api.raml
+++ b/api.raml
@@ -573,7 +573,7 @@ types:
     description: |
       Initiate a password reset.
       Sends reset email to primary email address with a link like
-      `https://idp-pw.local/#/recovery/verify/{resetUid}/{code}`
+      `https://idp-pw.local/#/recovery/verify/{resetUuid}/0`
       Responds with id of reset object and all available email addresses
       for the user.
     body:
@@ -591,7 +591,7 @@ types:
     /validate:
       put:
         securedBy: null
-        description: Validate reset code to complete password reset process
+        description: Validate reset UUID to complete password reset process
         body:
           properties:
             code: string
@@ -599,9 +599,6 @@ types:
           200:
             description: >
               Reset was validated. Subsequent API calls will use a http only cookie for the access token.
-          400:
-            description: >
-              The provided code was incorrect.
           410:
             description: >
               The code has expired.
--- a/api.raml
+++ b/api.raml
@@ -573,7 +573,7 @@ types:
     description: |
       Initiate a password reset.
       Sends reset email to primary email address with a link like
-      `https://idp-pw.local/#/recovery/verify/{resetUuid}/0`
+      `https://idp-pw.local/#/recovery/verify/{resetUid}/{code}`
       Responds with id of reset object and all available email addresses
       for the user.
     body:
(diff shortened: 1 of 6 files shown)


application/frontend/config/main.php
```diff
@@ -132,14 +132,9 @@
                 /*
                  * Reset routes
                  */
-                'GET /reset/' . UID_ROUTE_PATTERN => 'reset/view',
                 'POST /reset' => 'reset/create',
-                'PUT /reset/' . UID_ROUTE_PATTERN => 'reset/update',
-                'PUT /reset/' . UID_ROUTE_PATTERN . '/resend' => 'reset/resend',
```
mtompset, line 138: Why removing resend?

devon-sil:
@briskt With the recent change to default to "Do not disclose," one item we haven't addressed yet is how to handle if a person doesn't have access to any recovery methods.  Previously, there was a choice to email the manager on file, though that is not available anymore via the UI.  Is this change going to impact any ability to still utilize that? (be it through an API call, or perhaps an adjustment to the Do Not Disclose process to have one more step they could do to try to recover a password "if their account exists")

application/frontend/config/main.php
briskt: > Removed the GET /reset/{uid} and PUT /reset/{uid}/resend endpoints. These are no longer in use due the simpler reset flow. There is no use in retrieving the reset as there is no useful metadata available, and a re-send is done by repeating the POST /reset request.

briskt:
> @briskt With the recent change to default to "Do not disclose," one item we haven't addressed yet is how to handle if a person doesn't have access to any recovery methods. Previously, there was a choice to email the manager on file, though that is not available anymore via the UI. Is this change going to impact any ability to still utilize that? (be it through an API call, or perhaps an adjustment to the Do Not Disclose process to have one more step they could do to try to recover a password "if their account exists")

As discussed on [Chat](https://chat.google.com/room/AAQALXUKjiw/457K8ljvST0/457K8ljvST0?cls=10), when a user has no verified password reset options, their manager email address will be used. It's also worth noting that most people do have at least one because their personal email address is added automatically.

application/tests/api/ResetCest.php
```diff
@@ -41,85 +41,6 @@ public function test14(ApiTester $I)
...
-        $I->wantTo('check response when making PUT request for updating a reset object method supervisor');
-        $I->sendPUT('/reset/11111111111111111111111111111111', [
-            'type' => 'supervisor',
-        ]);
-        $I->seeResponseCodeIs(200);
-    }
-
-    public function test62(ApiTester $I)
-    {
-        $I->wantTo('check response when making authenticated DELETE request to /reset/id');
```
forevermatt, line 95: Do we no longer have a use-case for enabling a password reset request to be deleted?

forevermatt: Hmm... we didn't seem to have supported a `DELETE /reset/id` call before, either (based on the lack of an `actionDelete()` method in the controller and the lack of URL mapping for `DELETE` calls in the config). I'm confused now as to whether these tests would/should have been passing before this change, in addition to whether we should have tests for such a use case going forward. 😕

application/tests/unit/common/models/ResetTest.php
forevermatt, line 1: Will/Has some test been added (either here or in ID Broker) ensuring that we do fall back to sending the reset email to the supervisor if there are no other options available?

api.raml
forevermatt, line 578: This description seems to be out of date with the changes in this (or a subsequent?) PR.

application/tests/unit/common/models/ResetTest.php
briskt: Yes, in idp-id-broker reset-unit-tests.feature there are a few different scenarios about this.

api.raml
briskt: Thanks, I'll fix this. But I'll do so in another PR so as to not create merge issues.

application/tests/api/ResetCest.php
briskt: This was correct, actually. If the requester is authenticated, they would have seen a 405 (Method not allowed). If NOT authenticated, they would have seen a 401 (Unauthorized). I perhaps could have left them but didn't like how it was potentially misleading. (Not to mention the question of why PATCH and POST aren't tested if the idea is to have complete coverage.)

application/common/models/Reset.php
```diff
@@ -155,11 +131,6 @@ public function send(): void
...
         $this->trackAttempt('send');
 
-        if ($this->type === self::TYPE_SUPERVISOR) {
-            $this->sendSupervisor();
-            return;
-        }
-
         $methods = Method::getVerifiedMethods($this->user->employee_id);
         if (empty($methods) && $this->user->hasSupervisor()) {
             $this->sendPrimary();
```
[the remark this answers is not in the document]
briskt: Good idea, that's how it will be done in idp-id-broker. And this "send to all" is not new behavior.

api.raml
briskt: Included in #367

PR #365, opened 2026-05-18
Release 10.0.0-pre.2 - use Reset endpoints in idp-id-broker

briskt:
[IDP-1968](https://support.gtis.sil.org/issue/IDP-1968) remove reset table from idp-pw-api

---

### Changed (BREAKING)
- Use the password reset endpoints in idp-id-broker version 8.5.0 (pending release) rather than the local table.

---

### PR Checklist

- [x] Run `make composershow`

### DIFF
--- a/api.raml
+++ b/api.raml
@@ -573,7 +573,7 @@ types:
     description: |
       Initiate a password reset.
       Sends reset email to primary email address with a link like
-      `https://idp-pw.local/#/recovery/verify/{resetUid}/{code}`
+      `https://idp-pw.local/#/recovery/verify/{resetUuid}/0`
       Responds with id of reset object and all available email addresses
       for the user.
     body:
@@ -591,7 +591,7 @@ types:
     /validate:
       put:
         securedBy: null
-        description: Validate reset code to complete password reset process
+        description: Validate reset UUID to complete password reset process
         body:
           properties:
             code: string
@@ -599,9 +599,6 @@ types:
           200:
             description: >
               Reset was validated. Subsequent API calls will use a http only cookie for the access token.
-          400:
-            description: >
-              The provided code was incorrect.
           410:
             description: >
               The code has expired.
--- a/application/frontend/config/main.php
+++ b/application/frontend/config/main.php
@@ -133,9 +133,9 @@ return [
                  * Reset routes
                  */
                 'POST /reset' => 'reset/create',
-                'PUT /reset/' . UID_ROUTE_PATTERN . '/validate' => 'reset/validate',
+                'PUT /reset/<uuid>/validate' => 'reset/validate',
                 'OPTIONS /reset' => 'reset/options',
-                'OPTIONS /reset/' . UID_ROUTE_PATTERN . '/validate' => 'reset/options',
+                'OPTIONS /reset/<uuid>/validate' => 'reset/options',
 
                 /*
                  * User  routes
--- a/application/frontend/controllers/ResetController.php
+++ b/application/frontend/controllers/ResetController.php
@@ -4,19 +4,41 @@ namespace frontend\controllers;
 
 use common\components\personnel\NotFoundException;
 use common\helpers\Utils;
-use common\models\Reset;
 use common\models\User;
+use Exception;
 use frontend\components\BaseRestController;
+use Sil\Idp\IdBroker\Client\IdBrokerClient;
+use Sil\Idp\IdBroker\Client\ServiceException;
+use Yii;
 use yii\filters\AccessControl;
 use yii\helpers\ArrayHelper;
-use yii\helpers\Json;
 use yii\web\BadRequestHttpException;
 use yii\web\HttpException;
-use yii\web\NotFoundHttpException;
-use yii\web\ServerErrorHttpException;
 
 class ResetController extends BaseRestController
 {
+    public IdBrokerClient $idBrokerClient;
+
+    /**
+     * @throws Exception
+     */
+    public function init(): void
+    {
+        parent::init();
+        $config = Yii::$app->params['idBrokerConfig'];
+        if (empty($config['baseUrl']) || empty($config['accessToken'])) {
+            throw new Exception('ID_BROKER_baseURL and ID_BROKER_accessToken are required', 1778752282);
+        }
+        $this->idBrokerClient = new IdBrokerClient(
+            $config['baseUrl'],
+            $config['accessToken'],
+            [
+                IdBrokerClient::TRUSTED_IPS_CONFIG => $config['validIpRanges'] ?? [],
+                IdBrokerClient::ASSERT_VALID_BROKER_IP_CONFIG => $config['assertValidBrokerIp'] ?? true,
+            ]
+        );
+    }
+
     /**
      * Access Control Filter
      * NEEDS TO BE UPDATED FOR EVERY ACTION
@@ -40,19 +62,19 @@ class ResetController extends BaseRestController
     }
 
     /**
-     * Create new reset process
+     * Create new password reset. Calls the ID Broker to request a new Reset record and send one or more email
+     * messages containing a link to click, which will direct the user to the Validate action in this controller.
      * @return void
      * @throws BadRequestHttpException
-     * @throws NotFoundHttpException
-     * @throws ServerErrorHttpException
+     * @throws ServiceException
      */
     public function actionCreate(): void
     {
-        $username = trim(\Yii::$app->request->getBodyParam('username', ''));
-        $verificationToken = trim(\Yii::$app->request->getBodyParam('verification_token', ''));
+        $username = trim(Yii::$app->request->getBodyParam('username', ''));
+        $verificationToken = trim(Yii::$app->request->getBodyParam('verification_token', ''));
 
         if ($username === '') {
-            throw new BadRequestHttpException(\Yii::t('app', 'Reset.MissingUsername'));
+            throw new BadRequestHttpException(Yii::t('app', 'Reset.MissingUsername'));
         }
 
         /*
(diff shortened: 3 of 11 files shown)


application/frontend/config/main.php
```diff
@@ -133,9 +133,9 @@
                  * Reset routes
                  */
                 'POST /reset' => 'reset/create',
-                'PUT /reset/' . UID_ROUTE_PATTERN . '/validate' => 'reset/validate',
+                'PUT /reset/<uuid>/validate' => 'reset/validate',
                 'OPTIONS /reset' => 'reset/options',
-                'OPTIONS /reset/' . UID_ROUTE_PATTERN . '/validate' => 'reset/options',
+                'OPTIONS /reset/<uuid>/validate' => 'reset/options',
```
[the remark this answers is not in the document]
forevermatt: I think this switch to `uuid` is fine, especially if we're only internally using it as a UUID.

On a related note, I don't think the `UID_ROUTE_PATTERN` was correct, as it matched exactly 32 characters but a UUID is 32 characters _plus_ the hyphens, so 36 characters total.

application/frontend/controllers/ResetController.php
```diff
@@ -40,173 +62,67 @@ public function behaviors()
...
-     * @throws \Throwable
+     * Validate reset code. If successful, a limited-access cookie will be set.
+     * @param string $uuid
+     * @return void
+     * @throws HttpException
+     * @throws NotFoundException
+     * @throws Exception
      */
-    public function actionValidate($uid)
+    public function actionValidate(string $uuid): void
```
forevermatt, line 109: It looks like this function merely "verifies" a password reset request by sending in the UUID from the URL, without requiring any kind of additional code to be entered. Is that the intended behavior?

application/frontend/controllers/ResetController.php
```diff
@@ -40,173 +62,67 @@ public function behaviors()
...
     {
-        /** @var Reset $reset */
-        $reset = Reset::findOne(['uid' => $uid]);
-        if ($reset === null) {
-            throw new NotFoundHttpException();
+        try {
+            $brokerResponse = $this->idBrokerClient->verifyReset($uuid);
```
forevermatt, line 112: When the ID Broker API is called to verify a reset request, does ID Broker automatically delete the reset request if it is valid? (We could see IDP PW API do so in the old code here, but it's not obvious from the new code here whether it is deleted from ID Broker at that point now).

application/tests/api/ResetCest.php
```diff
@@ -41,56 +41,36 @@ public function test14(ApiTester $I)
...
-        $I->wantTo('check response when making multiple authenticated PUT request to validate a reset code');
-        for ($i = 0; $i < 10; $i++) {
-            $I->sendPUT('/reset/33333333333333333333333333333334/validate', ['code' => '344']);
-            if ($i < 9) {
-                $I->seeResponseCodeIs(400);
-            } else {
-                $I->seeResponseCodeIs(429);
-            }
-        }
-    }
```
forevermatt, line 92: Was a rate-limit test similar to this added to the ID Broker API?

application/tests/api/ResetCest.php
```diff
@@ -41,56 +41,36 @@ public function test14(ApiTester $I)
...
+        $I->sendPUT('/reset/11111111111111111111111111111111/validate');
         $I->seeResponseCodeIs(200);
     }
 
-    public function test81(ApiTester $I)
+    public function test810(ApiTester $I)
     {
         $I->wantTo('check response when making unauthenticated PUT request to validate an expired reset code');
-        $I->sendPUT('/reset/33333333333333333333333333333334/validate', ['code' => '444']);
+        $I->sendPUT('/reset/22222222222222222222222222222222/validate');
```
forevermatt, line 54: This is one of the places where it seems like Gherkin-style tests have an advantage over tests that depend on fixtures: This code depends on a different file having set that `22222...` reset to be expired, rather than this test itself ensuring that precondition is met.

Not a showstopper at all (I'm fine with proceeding with this code). I just wanted to highlight this as an example of the problem of fixtures "hiding" relevant data.

api.raml
```diff
@@ -573,7 +573,7 @@ types:
     description: |
       Initiate a password reset.
       Sends reset email to primary email address with a link like
-      `https://idp-pw.local/#/recovery/verify/{resetUid}/{code}`
+      `https://idp-pw.local/#/recovery/verify/{resetUuid}/0`
```
forevermatt, line 576: Why is there a `0` on the end of this URL?

api.raml
```diff
@@ -591,17 +591,14 @@ types:
     /validate:
       put:
         securedBy: null
-        description: Validate reset code to complete password reset process
+        description: Validate reset UUID to complete password reset process
```
forevermatt, line 594: I'm a little concerned with the removal of the code that had to be entered manually. Now all of the data necessary for doing a password reset seems to be in the URL, which is less protected than data included in headers, a POST body, or something that has to be entered _after_ arriving at the specified URL.

@briskt Thoughts? I might just not yet know/understand some applicable detail of the new design.

api.raml
```diff
@@ -591,17 +591,14 @@ types:
...
         body:
           properties:
             code: string
         responses:
```
forevermatt, line 598: What does this API endpoint now return for an unknown UUID (which I think is now the only form of an invalid "code", since code was removed)?

compose.yaml
```diff
@@ -155,7 +155,7 @@ services:
       PMA_PASSWORD: broker
 
   broker:
-    image: ghcr.io/sil-org/idp-id-broker:8
+    image: ghcr.io/sil-org/idp-id-broker:verify-reset
     platform: linux/amd64
     ports:
```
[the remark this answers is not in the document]
forevermatt: I assume this is for testing during pre-release, which is great. 👍

application/frontend/controllers/ResetController.php
```diff
@@ -40,173 +62,67 @@ public function behaviors()
...
+        } catch (ServiceException $e) {
+            throw new HttpException($e->httpStatusCode);
```
forevermatt, line 114: This direct-passing of the ID Broker API call's HTTP status code back as the HTTP Status Code for this ID PW API call means we aren't directly controlling what all HTTP status codes this IDP PW API endpoint can return (e.g. only those documented in our RAML file).

I don't know that that's a bad thing, but it seems like sort of a non-obvious thing, and one that will make it easy for us to miss (and therefore not document) HTTP Status Codes that this API endpoint might now return.

Did you have a particular reason for doing it this way, or was it just a nice simple way to handle the various possible status codes?

application/frontend/config/main.php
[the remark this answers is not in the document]
briskt: > Consider adding a regex constraint for <uuid> ... so only valid reset UUID formats route to this action.

I think this isn't necessary. There are no other routes that potentially conflict. Making it unnecessarily restrictive is generally confusing when debugging.

> I think this switch to uuid is fine, especially if we're only internally using it as a UUID.

@forevermatt I am not sure I follow. Why would it _not_ be fine if we used it other than internally?

application/frontend/controllers/ResetController.php
briskt: It was an intentional decision to simplify and "get out of the way" in a sense. The goal is to make idp-pw-api dirt simple so that id-broker is the single source of truth wherever possible.

application/frontend/controllers/ResetController.php
```diff
@@ -40,173 +62,67 @@ public function behaviors()
     }
 
     /**
-     * Create new reset process
+     * Create new password reset. Calls the ID Broker to request a new Reset record and send one or more email
+     * messages containing a link to click, which will direct the user to the Validate action in this controller.
```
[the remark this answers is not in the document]
briskt: Rather than "except", I find that "optional" behaves more appropriately. The "except" keyword made it impossible to work with both authenticated an unauthenticated requests.

application/frontend/controllers/ResetController.php
```diff
@@ -40,173 +62,67 @@ public function behaviors()
...
-                $reset->restart();
-            } else {
-                $reset->send();
-            }
-        }
+        $this->idBrokerClient->createReset($username);
 
-        \Yii::$app->response->statusCode = 204;
+        Yii::$app->response->statusCode = 204;
     }
```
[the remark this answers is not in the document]
briskt: OK, good catch. Addressed in PR #368.

application/frontend/controllers/ResetController.php
briskt: Yes, it is. I'll say more on your other comment about this.

application/tests/api/ResetCest.php
```diff
@@ -41,56 +41,36 @@ public function test14(ApiTester $I)
...
     {
         $I->wantTo('check response when making unauthenticated PUT request to validate a reset code');
-        $I->sendPUT('/reset/33333333333333333333333333333333/validate', ['code' => '333']);
+        $I->sendPUT('/reset/11111111111111111111111111111111/validate');
         $I->seeResponseCodeIs(200);
     }
 
-    public function test81(ApiTester $I)
+    public function test810(ApiTester $I)
     {
```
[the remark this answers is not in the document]
briskt: There is no pattern that I've been able to decipher. A separate PR would be best for sorting out the clutter.

application/tests/api/ResetCest.php
briskt: I agree except for the part that Gherkin is the only or best way to ensure that the test sets up the preconditions in a readable way. These idp-pw-api tests are a nightmare to debug, but I think they could be overhauled to locate the test fixture setup right here.

application/tests/api/ResetCest.php
briskt: No. It's not applicable with the new protocol. Either a requester has the UUID or they don't. If a bot tries a list of UUID values, we can't block that at the application layer. To be honest, I don't know how the existing rate limit algorithm helps. If a bot somehow guesses the UID, they could try to brute-force the numeric code, but how would they get the UID in the first place without also getting the code?

api.raml
briskt: For backward compatibility so the idp-profile-ui code doesn't have to change.

api.raml
briskt: The numeric code is worthless. The UUID and code always exist together.
I hadn't thought of the difference between data in the URL and data in the request body, but even that offers no benefit since the UUID becomes invalid when it's used the first time. Since the only place an attacker could get the UUID but not the numeric code is by snooping on the URLs in idp-pw-api or idp-id-broker request traffic, at the moment they see it go through it becomes invalid. Maybe I missed a different attack vector. What do you think?

api.raml
briskt: It returns a 404 for either an invalid UUID or an expired reset record. The RAML file is updated to reflect this in the next PR (#366).

api.raml
```diff
@@ -591,17 +591,14 @@ types:
...
           200:
             description: >
               Reset was validated. Subsequent API calls will use a http only cookie for the access token.
-          400:
-            description: >
-              The provided code was incorrect.
           410:
             description: >
               The code has expired.
```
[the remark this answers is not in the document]
briskt: The next PR (#366) fixed this.

compose.yaml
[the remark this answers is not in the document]
briskt: Yes, I just released 8.5.0 today, so I'll set this back to `8`. Addressed in PR #368.

application/frontend/controllers/ResetController.php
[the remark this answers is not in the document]
briskt: Addressed in PR #368.

application/frontend/controllers/ResetController.php
briskt: It immediately expires it, so it's rendered ineffective and will be deleted during maintenance.

application/tests/api/ResetCest.php
```diff
@@ -41,56 +41,36 @@ public function test14(ApiTester $I)
...
         $I->wantTo('check response when making unauthenticated PUT request to validate an expired reset code');
-        $I->sendPUT('/reset/33333333333333333333333333333334/validate', ['code' => '444']);
+        $I->sendPUT('/reset/22222222222222222222222222222222/validate');
         $I->seeResponseCodeIs(410);
```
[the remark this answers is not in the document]
briskt: Addressed in PR #368.

application/tests/api/ResetCest.php
```diff
@@ -41,56 +41,36 @@ public function test14(ApiTester $I)
...
     }
 
-    public function test83(ApiTester $I)
+    public function test830(ApiTester $I)
     {
         $I->wantTo('check response when making authenticated DELETE request to reset/id/validate');
         $I->setCookie('access_token', 'user1', parent::getCookieConfig());
-        $I->sendDELETE('/reset/33333333333333333333333333333333/validate', ['code' => '333']);
+        $I->sendDELETE('/reset/44444444444444444444444444444444/validate');
         $I->seeResponseCodeIs(405);
```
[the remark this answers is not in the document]
briskt: Addressed in PR #368.

application/frontend/config/main.php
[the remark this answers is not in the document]
forevermatt: Sorry, my intended emphasis was on using it only as a UUID, not using it only internally.

application/tests/api/ResetCest.php
forevermatt: Okay, great. 👍  I perhaps have a misunderstanding of the word "fixture" then, as I thought it referred to data defined somewhere else that is put in place before the tests run, rather than some data that a test itself puts in place as part of it's "Given" (re: Given/When/Then) or "Arrange" (re: Arrange/Act/Assert) setup.

You also brought up a good point that I tend to conflate Gherkin-style testing with being clear about test setup, the action being tested, and verification of the results. Both have value, but Gherkin isn't the only testing approach that does that. (I think I really like Gherkin because it provides _both_ a clear arrange/act/assert structure _and_ it makes the the tests so easy to read.)

Thanks for being willing to continue discussing this topic. 🙂  Hopefully I'm not getting to be too frustrating in my repetitious comments trying to push us towards easier-to-understand tests.

briskt: That's interesting. I hadn't thought of a strict definition of "fixture" like that. To seek a better understanding, I asked Gemini. Rather than copying the entire thread, here are the prompts I used:

> What is an accurate definition of fixture in the context of automated software testing?

> Is there anything in a textbook definition that says whether it is defined once for a suite of tests or separately for each test or test case?

A very brief summary is that it can be both. But the full answer was very interesting to read.

As for the Gherkin commentary, no one could argue that a Gherkin feature is easy to read, assuming that it was written well and the reader's native language is English. One of my complaints is that it's so verbose I lose patience in reading the natural language text. Another is that you don't get to see what the code does unless you hop back and forth between the feature file and the context file. That's especially problematic if you're not in an IDE or your IDE is broken (If you have any tips for fixing this, I'd very much appreciate your help.)

> Thanks for being willing to continue discussing this topic.

Are you kidding?! In case you hadn't noticed, I love to debate! And I really appreciate your eye for quality.

application/frontend/config/main.php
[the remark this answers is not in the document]
briskt: OK

application/tests/api/ResetCest.php
forevermatt: > One of my complaints is that it's so verbose I lose patience in reading the natural language text. Another is that you don't get to see what the code does unless you hop back and forth between the feature file and the context file. That's especially problematic if you're not in an IDE or your IDE is broken (If you have any tips for fixing this, I'd very much appreciate your help.)

Let's definitely talk about this at some point. To my ear, "verbose" is itself a code smell for tests (including those written using Gherkin), and I think it's just a poor-quality habit that's too easy to fall into with Gherkin/Behat style tests because of the temptation to reuse several existing steps, rather than come up with a single, clear step that states at the appropriate level what really needs to be true for that particular test. And the issue of needing to hop back and forth I think reflects a difference in how you think about / review these tests than I do. I may have something to learn here. 🙂 

If your IDE doesn't let you control-click on the Gherkin step and be taken straight to the Context function called for it, that does sound like simply an IDE-configuration problem we could hopefully get fixed.

api.raml
forevermatt: I still don't really understand why the `0` needs to be there, but it's probably not worth spending more of your time on. Thanks, though.

api.raml
forevermatt: I was thinking of a dictionary attack, where a bad actor simply iterates through possible URLs until it finds one that works (which is also possibly why I tend to think in terms of rate-limits).

api.raml
briskt: This is the describing the URL in the email being sent to the user. Since it's confusing and unnecessary in this document, I'll remove it.

PR #377, opened 2026-06-01
PR 2 of 3: Access-token flow – update AuthController logout to use IdBroker

briskt:
[IDP-1967](https://support.gtis.sil.org/issue/IDP-1967) move access token storage to idp-id-broker

---

### Changed
- Access-token flow – update AuthController logout to use IdBroker lookup.

---

### PR Checklist

- [ ] Put version number in PR title (e.g. `Release x.y.z - Summary of changes`)
- [ ] Unit tests created or updated
- [ ] Run `make composershow`

### DIFF
--- a/application/frontend/controllers/AuthController.php
+++ b/application/frontend/controllers/AuthController.php
@@ -2,9 +2,9 @@
 
 namespace frontend\controllers;
 
+use common\components\auth\AuthnInterface;
 use common\components\auth\RedirectException;
 use common\components\auth\User as AuthUser;
-use common\components\auth\AuthnInterface;
 use common\components\personnel\NotFoundException;
 use common\helpers\Utils;
 use common\models\User;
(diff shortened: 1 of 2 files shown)


application/frontend/controllers/AuthController.php
```diff
@@ -117,17 +117,24 @@ public function actionLogout()
...
+            /*
+             * Look up and clear the token in IdBroker, then log out of IdP
+             */
+            try {
+                $accessTokenHash = Utils::getAccessTokenHash($accessToken);
+                /** @var \common\components\personnel\PersonnelInterface $personnel */
+                $personnel = \Yii::$app->personnel;
+                $personnelUser = $personnel->findByAccessToken($accessTokenHash);
+
+                $personnel->clearAccessToken($personnelUser->employeeId);
```
hobbitronics, line 140: It seems a little counter intuitive that the comment was renamed to 'Remove' but the method used here is 'Clear' which was the previous comment. Probably fine though.

application/frontend/controllers/AuthController.php
```diff
@@ -117,17 +117,24 @@ public function actionLogout()
...
 
                 /** @var AuthUser $authUser */
-                $authUser = $user->getAuthUser();
+                $authUser = User::createFromPersonnelUser($personnelUser)->getAuthUser();
```
hobbitronics: I feel like this method name is misleading. I think construct is a better verb than create as that makes me think it's created in the db.

hobbitronics (approved):
Code looks good, but I may not understand the nuances of the change. Does the `User::findOne` method use the idBroker and that's why it's not being used anymore?

application/frontend/controllers/AuthController.php
briskt: Yeah, very strange that Copilot decided to change that comment. It is a comment about the browser, though, but I'm still not sure why it did that. Maybe to distinguish broker and browser? "clear from the broker and remove from the browser"

forevermatt (approved):

briskt:
Note: tests are not passing because this is incomplete without PR 3 in this series.

PR #395, opened 2026-07-24
allow for multiple frontend origins

briskt:
[IDP-2266](https://support.gtis.sil.org/issue/IDP-2266) Add multi-origin support to idp-pw-api

---

### Added
- Added a new configuration option (`TRUSTED_ORIGINS`) to provide a list of trusted origins, replacing `UI_URL` and `UI_CORS_ORIGIN`. This will facilitate a domain name change and possibly a multi-tenant architecture further down the road.

### Fixed
- Added `COMPOSER_ROOT_VERSION` to other test containers to silence the message: "Composer could not detect the root package (sil-org/idp-pw-api) version, defaulting to '1.0.0'. See https://getcomposer.org/root-version"
- Changed `getCookieConfig()` visibility to `protected` to keep it from being run as a test.
- Added `ext-bcmath` to composer.json to reflect the change in #393 

### Deprecated
- Deprecated `UI_URL` and `UI_CORS_ORIGIN` in favor of the new `TRUSTED_ORIGINS`.

---

### PR Checklist

- [ ] Put version number in PR title (e.g. `Release x.y.z - Summary of changes`)
- [x] Documentation (README, openapi.yaml, local.env.dist, etc.)
- [ ] Run `make composershow`


application/frontend/controllers/AuthController.php
```diff
@@ -150,28 +150,29 @@ public function actionLogout()
...
                 $auth = \Yii::$app->auth;
-                $auth->logout(\Yii::$app->params['uiUrl'], $authUser);
+                $auth->logout(Utils::getTrustedUiUrl(), $authUser);
             } catch (RedirectException $e) {
                 return $this->redirect($e->getUrl());
             }
         }
 
-        return $this->redirect(\Yii::$app->params['uiUrl']);
+        return $this->redirect(Utils::getTrustedUiUrl());
```
hobbitronics, line 159: Would it make sense to get this once instead of calling it in several different places?

application/tests/api/CorsCest.php
```diff
@@ -0,0 +1,29 @@
...
+use BaseCest;
+
+require_once "BaseCest.php";
+
+class CorsCest extends BaseCest
+{
+    public function testTrusted(ApiTester $I)
+    {
+        $I->wantTo('check response when making a request from a trusted origin');
+        $I->haveHttpHeader('Origin', 'http://localhost');
```
hobbitronics, line 15: Would it be worth testing the second trusted origin as well? What about testing when none is set?

hobbitronics (approved):
Looks good. I asked Copilot if we're missing any edge cases. 1 and 2 seem worthwile and easy enough.

1. a second trusted origin (e.g. https://localhost) to prove multi-origin support is actually working, not just the old single-origin case
2. a request with no Origin header
3. Since this PR changes redirect/origin selection via Utils::getTrustedUiUrl(), I wonder if we should add an auth/logout-oriented test for the untrusted or missing-origin case as well, so the resulting redirect target is explicit and protected against regressions.

briskt:
This does not work. It breaks the login flow because not all requests have an origin header.

PR #400, opened 2026-07-29
update dependencies.json

briskt:
### Fixed
- Updated the dependencies.json file to reflect recent dependabot changes.

### DIFF
--- a/application/dependencies.json
+++ b/application/dependencies.json
@@ -133,7 +133,7 @@
   },
   {
     "name": "guzzlehttp/guzzle",
-    "version": "7.11.1"
+    "version": "7.15.1"
   },
   {
     "name": "guzzlehttp/guzzle-services",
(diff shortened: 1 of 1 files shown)


application/dependencies.json
```diff
@@ -447,9 +447,21 @@
...
     "version": "2.0.4"
   },
+  {
+    "name": "simplesamlphp/assert",
+    "version": "v2.0.2"
+  },
+  {
+    "name": "simplesamlphp/composer-xmlprovider-installer",
+    "version": "v1.3.1"
+  },
```
hobbitronics, line 457: why are these added? Just making them explicit?

hobbitronics (approved):

application/dependencies.json
briskt: I think they're required in a new version of simplesamlphp but I'm not sure about how composer works.

